PR #53, opened 2023-11-02
[FIX] vscode: fix addons detection irrelevant files in odoo path

livaios:


package-lock.json
fda-odoo, line 1: Ha, I think we can add package_lock in the gitignore file :)

vscode/client/extension.ts
```diff
@@ -327,36 +327,39 @@ function deleteOldFiles(context: ExtensionContext, outputChannel: OutputChannel)
...
+                    break
+            }
         }
     }
 }
 
 async function checkOdooPath(context: ExtensionContext) {
     let currentConfig = getCurrentConfig(context)
-    let odooFound = workspace.getWorkspaceFolder(Uri.parse(currentConfig.odooPath))
+    let odooFound = currentConfig ? workspace.getWorkspaceFolder(Uri.parse(currentConfig.odooPath)) : true
```
fda-odoo, line 362: Why true if do not find the currentConfig?

vscode/client/extension.ts
```diff
@@ -327,36 +327,39 @@ function deleteOldFiles(context: ExtensionContext, outputChannel: OutputChannel)
...
-            case ("Update current configuration"):
-                ConfigurationWebView.render(context, currentConfig.id);
-                break
-            case ("View Paths"):
-                odooOutputChannel.show();
-                break
+    if (currentConfig) {
+        let missingFiles = files.filter(file => {
+            return !currentConfig.addons.some((addon) =>
+                file.fsPath.startsWith(addon) || file.fsPath.startsWith(currentConfig.odooPath)
```
fda-odoo, line 333: Why `file.fsPath.startsWith(currentConfig.odooPath)` is in the `some` but is not using addon?
`return !currentConfig.addons.some((addon) =>file.fsPath.startsWith(addon)) || !file.fsPath.startsWith(currentConfig.odooPath` is maybe a little more readable

fda-odoo (changes requested):

package-lock.json
livaios: Ahh my bad I thought it was already there

vscode/client/extension.ts
livaios: not finding the currentConfig means that the extension is disabled so we will not check for the odoo path to match

PR #71, opened 2023-11-15
[IMP] vscode, server: new config to fine tune the missing import messages

fda-odoo:


server/core/import_resolver.py
```diff
@@ -91,7 +95,9 @@ def _resolve_packages(file_symbol, level, from_stmt):
     return file_tree
 
 def _get_or_create_symbol(ls, symbol, names, file_symbol, asname, start_pos, end_pos):
-    """try to return sub symbol that is a file or package, or create the symbol"""
+    """try to return sub symbol that is a file or package, or create the symbol. The second returned element is the last
+    successfully loaded symbol in the search."""
+    last_symbol = symbol
```
Louciole, line 100: Why are you doing that? Is this assignment used somewhere?

Louciole (commented):
Aside my two comments it look good !

server/core/import_resolver.py
fda-odoo: in the return statement :) It is used to know until which point the search has been, so we can identify which kind of import it is

server/core/python_arch_eval.py
```diff
@@ -69,18 +69,28 @@ def visit_ImportFrom(self, node):
         self.eval_symbols_from_import_stmt(node.module,
                     node.names, node.level, node)
 
+    def _match_diag_config(self, symbol):
+        import_diag_level = Odoo.get().config.diag_missing_imports
+        if import_diag_level == "none":
+            return False
+        if import_diag_level == "all":
+            return True
+        if import_diag_level == "only_odoo" and symbol and symbol.get_tree() and symbol.get_tree()[0] and symbol.get_tree()[0][0] == "odoo":
```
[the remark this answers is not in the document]
Louciole: Fair enough, I don't see a better way neither

PR #85, opened 2023-11-22
[IMP] Client side code cleaning and Python extension integration

### LINKED ISSUE #78: Odoo-ls not killed automatically when I kill VS Code
**Your setup**
Version: 0.2.1 &0.2.2
Operating System: Ubuntu 22.04 and i3 desktop manager
IDE and/or Integration tool (for example: Vscode - official extension): Vscode - official extension

**Describe the bug**
When I kill VS Code (Ctrl+q) the extension process is not killed automatically, annoying when I am working on multiples instances of VS code and each odoo-ls server is taking 1.8Go of RAM memory

**To Reproduce**
Steps to reproduce the behavior:*
Kill VS Code with Ctrl+q and check in the monitor that the instances of odoo-ls is still here 

**Expected behavior**
It should be killed at the same time as VS Code ?

delvsola:
This PR aims to provide a first fix to many errors in the client code and acts as an intermediate patch until the client side is properly refactored.

The extension also now integrates VSCode's Python extension, which means the Language Server will now be launched with the Interpreter selected for that workspace.

## Changes

- VSCode Python extension integration
- Removal of the Python Path field in the Configuration webview.
- Better logging in the Odoo channel.

## Fixes

- The Language Client is now properly stored and referenced through a correctly defined global variable.
- Server stops are now properly awaited, which should hopefully fix #78 .
- Configuration changes are now queued during server stops.
- LanguageClient is not disposed between Configuration changes anymore, only between interpreter changes.


vscode/client/extension.ts
```diff
@@ -742,21 +743,29 @@ async function checkStandalonePythonVersion(context: ExtensionContext): Promise<
...
-        const interpreter = await getInterpreterDetails();
+    try {
+        interpreter = await getInterpreterDetails();
+    } catch {
+        interpreter = null;
+    }
+
+    //trying to use the VScode python extension
+    if (interpreter && global.IS_PYTHON_EXTENSION_READY !== false) {
         config ? pythonPath = interpreter.path[0] : pythonPath = null;
```
delvsola, line 756: `pythonPath = config ? interpreter.path[0] : null`

delvsola:
@Louciole The extension crashes when starting as Disabled with the Python extension enabled

Louciole:
> @Louciole The extension crashes when starting as Disabled with the Python extension enabled

I can't reproduce it (either in the debug or packaged version) did you do something special?

vscode/client/common/python.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const onDidChangePythonInterpreterEvent = new EventEmitter<IInterpreterDetails>();
+export const onDidChangePythonInterpreter: Event<IInterpreterDetails> = onDidChangePythonInterpreterEvent.event;
+
+let _api: PythonExtension | undefined;
+async function getPythonExtensionAPI(): Promise<PythonExtension | undefined> {
+    if (_api) {
+        return _api;
+    }
+    _api = await PythonExtension.api();
```
fda-odoo, line 21: are we sure it will never become invalid?

Louciole: We only use this Promise right after getting it and getting it each time we need it to be as close to the current configuration as we can be so there's no reason for it to be invalid in the current setup (but maybe I didn't understand the comment)

PR #86, opened 2023-11-22
[IMP] server: dependencies data structure

fda-odoo:
change dependencies data structure from dictionnary to classes to lower the memory consumption (~5%)


fda-odoo:
not efficient enough

PR #88, opened 2023-11-29
[IMP] server: speedup exist_cs on windows

fda-odoo:
if the OS is case_preserving, the actual code is using Path.listdir to list list files in a directory. os.scandir is faster, and this commit uses this method instead


fda-odoo:
To test on windows before merge

fda-odoo:
Seems around 100% SLOWER. Safe to close without regrets

PR #100, opened 2024-01-03
[FIX] server: avoid index out of range on missing symbol search

fda-odoo:
When searching for missing path after creating a file, if the path was found in the not_found_paths, the server crashed with a index not found error.


Louciole (approved):
it's the most beautiful while i've seen this year!

PR #130, opened 2024-08-29
[IMP] server: refactor SyncOdoo.pop_item

mmahrouss:
Refactoring changes on SyncOdoo.pop_item, but no change in logic

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -351,14 +351,11 @@ impl SyncOdoo {
         let mut arc_sym: Option<Rc<RefCell<Symbol>>> = None;
         //Part 1: Find the symbol with a unmutable set
         {
-            let set =  if step == BuildSteps::ARCH_EVAL {
-                &self.rebuild_arch_eval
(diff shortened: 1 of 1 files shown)


server/src/core/odoo.rs
```diff
@@ -367,34 +364,15 @@ impl SyncOdoo {
...
-                            }
-                        }
-                    }
+                    let index_set =  match index {
+                        x if x == BuildSteps::ARCH as usize => &self.rebuild_arch,
+                        x if x == BuildSteps::ARCH_EVAL as usize => &self.rebuild_arch_eval,
+                        x if x == BuildSteps::VALIDATION as usize => &self.rebuild_validation,
+                        _ => continue,
+                    };
+                    dep_set.iter()
```
fda-odoo, line 373: I do not really like the for_each :/
Why not a

```
current_count += dep_set.iter().filter(...).count();
```

mmahrouss: That looks good too, I will add it

PR #142, opened 2024-10-18
[FIX] server: Avoid adding odoo and addons paths in syspath

mmahrouss:


server/src/core/odoo.rs
```diff
@@ -171,6 +171,9 @@ impl SyncOdoo {
                     let pathbuf = PathBuf::from(path);
                     if pathbuf.is_dir() {
                         let final_path = pathbuf.sanitize();
+                        if final_path == session.sync_odoo.config.odoo_path || session.sync_odoo.config.addons.contains(&final_path) {
```
fda-odoo, line 174: It will work in most cases, but then some folders are not added to root symbols if they are in odoo addons.
But if they are in sys.path, it means that the user probably want to import some things for this folder. So I would let the addition to root path, but not marking it as "is_external".
Of course it can make all odoo addons available for import without "odoo.addons" before, but that's the user choice

mmahrouss: Okay I understand, I will make a new version and let you know! 

Thanks for the feedback:)

PR #154, opened 2024-11-05
[IMP] server: use arena as memory allocator

fda-odoo:


fda-odoo:
Not useful anymore, the code has been refactored now :)

PR #158, opened 2024-11-15
[FIX] server: prevent parsing docstrings as markdown codeblocks

mmahrouss:


fda-odoo:
Shouldn't we remove space instead of transforming them into nbsp?

mmahrouss:
> Shouldn't we remove space instead of transforming them into nbsp? Spaces before comments are there only for the indentation of the documented block. But the documentation we show do not have to be indented

Yes that was my first approach. 

What was confusing was if I should always trim the spaces from the beginning of lines, that will make us lose some of the formatting of the docstring, right ? 

What I could do is remove the common number of spaces on the lines ? but that will not work in all cases tho? 

What do you think ?

mmahrouss:
@fda-odoo 

I have a version with common removed spaces, the code is a bit overkill, and I do not really like the result
image
image


As long as there is another indentation, it starts to have the same problem

fda-odoo:
Indeed, it's not ok. We still have the syntaxical colors... 
Maybe we should not use markdown for comments, but I don't know if we can create a block that do not use markdown :/

PR #160, opened 2024-11-26
[IMP] server: represent and evaluate super

mmahrouss:


server/src/core/evaluation.rs
```diff
@@ -562,27 +566,47 @@ impl Evaluation {
...
-                                    factory: None,
-                                    get_symbol_hook: None,
-                                },
-                                value: None,
-                                range: Some(expr.range)
-                            });
+                            if base_sym.borrow().get_tree() == (vec![S!("builtins")], vec![S!("super")]){
+                                match parent.borrow().get_in_parents(&vec![SymType::CLASS], true){
+                                    None => (), // TODO, diagnostic? or just leave empty
+                                    Some(parent_class) =>
```
fda-odoo, line 579: You only handle super(), and not super(Class) nor super(Class, instance), right?

fda-odoo, line 579: You only handle super(), and not super(Class) nor super(Class, instance), right?

server/src/core/symbols/symbol.rs
```diff
@@ -1618,35 +1622,37 @@ impl Symbol {
...
+                        } else {
+                            result.insert(name.clone(), vec![(symbol, None)]);
+                        }
                     }
                 }
                 let bases = symbol.borrow().as_class_sym().bases.clone();
                 for base in bases.iter() {
                     //no comodel as we will process only model in base class (overrided _name?)
-                    Symbol::all_members(&base, session, result, false, from_module.clone(), acc);
+                    Symbol::all_members(&base, session, result, false, from_module.clone(), acc, None);
```
mmahrouss, line 1639: for bases, don't we need to skip duplicate symbols by name ?

It only makes sense to have duplicate names in the all members for co-models, right? 
to discuss

PR #164, opened 2024-12-11
[IMP] server: avoid repetition in attribute completion

mmahrouss:


server/src/core/symbols/symbol.rs
```diff
@@ -1651,31 +1651,31 @@ impl Symbol {
...
                                 }
                             }
                         }
                     }
                 }
+                let bases = symbol.borrow().as_class_sym().bases.clone();
+                for base in bases.iter() {
+                    //no comodel as we will process only model in base class (overrided _name?)
+                    Symbol::all_members(&base, session, result, false, from_module.clone(), acc, false);
+                }
```
mmahrouss, line 1678: Here I just read the comodels first before the bases to get the ride order (mostly)

server/src/features/completion.rs
```diff
@@ -554,8 +554,8 @@ fn complete_attribut(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>, attr
                         Symbol::all_members(&parent_sym, session, &mut all_symbols, true, from_module, &mut None, parent_sym_eval_weak.is_super);
                         for (_symbol_name, symbols) in all_symbols {
                             //we could use symbol_name to remove duplicated names, but it would hide functions vs variables
-                            for (final_sym, dep) in symbols.iter() {
-                                if final_sym.borrow().name().starts_with(attr.attr.id.as_str()) {
+                            if _symbol_name.starts_with(attr.attr.id.as_str()) {
+                                if let Some((final_sym, dep)) = symbols.first() {
                                     items.push(build_completion_item_from_symbol(session, final_sym, dep.clone()));
                                 }
```
mmahrouss, line 560: I take the first from the vector, because it should be enough, right?

PR #167, opened 2024-12-19
Better domains - autocompletion, diagnostics, hover

fda-odoo:
The PR contains all changes and fixs about search domains.
- Autocompletion for fields
- comprehension of fields that contains dots
- fix wrong diagnostics of invalid fields
- hover of fields

### DIFF
--- a/server/src/features/completion.rs
+++ b/server/src/features/completion.rs
@@ -8,7 +8,7 @@ use tracing::info;
 use weak_table::traits::WeakElement;
 
 use crate::constants::SymType;
-use crate::core::evaluation::{ContextValue, Evaluation, EvaluationSymbolPtr, EvaluationSymbolWeak};
+use crate::core::evaluation::{Context, ContextValue, Evaluation, EvaluationSymbol, EvaluationSymbolPtr, EvaluationSymbolWeak};
 use crate::core::import_resolver;
 use crate::core::python_arch_eval_hooks::PythonArchEvalHooks;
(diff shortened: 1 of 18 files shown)


server/src/features/completion.rs
```diff
@@ -565,7 +575,48 @@ fn complete_string_literal(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>
...
+                        ..Default::default()
+                    });
+                }
+            },
+            ExpectedType::DOMAIN_FIELD(parent) => {
+                let mut all_symbols: HashMap<String, Vec<(Rc<RefCell<Symbol>>, Option<String>)>> = HashMap::new();
+                let from_module = file.borrow().find_module().clone();
+                Symbol::all_members(&parent, session, &mut all_symbols, true, from_module, &mut None, false);
+                for (_symbol_name, symbols) in all_symbols {
+                    //we could use symbol_name to remove duplicated names, but it would hide functions vs variables
```
mmahrouss, line 618: But we cannot have functions and variables with the same name, can we?

mmahrouss (approved):
Nice really cool! 
Just one question on a comment, otherwise looks great!

server/src/features/completion.rs
fda-odoo: We could. Some would be shadowed, but as we are searching here in classes, inheritance, co-models,... we could find different symbols type with the same name

PR #178, opened 2025-01-24
[FIX] server: avoid validating when file ast changes

mmahrouss:


server/src/core/python_validator.rs
```diff
@@ -89,9 +93,12 @@ impl PythonValidator {
                     let mut builder = PythonArchEval::new(func.clone());
                     builder.eval_arch(session);
                 }
+                if file_info_rc.borrow().text_hash != func.borrow().get_processed_text_hash(){
```
fda-odoo, line 96: As the Arch is rebuilt 6 lines earlier, we could rebuild the arch now if text_hash is invalid.
So maybe something more like:
```rust
if file_info_rc.borrow().text_hash != func.borrow().get_processed_text_hash() || func.borrow().as_func().arch_status == BuildStatus::PENDING {
    let mut builder = PythonArchBuilder::new(func.clone());
    builder.load_arch(session);
}
```

fda-odoo (changes requested):

server/src/core/python_validator.rs
mmahrouss: Didn't we want to make it trigger naturally? while we get the right message, on save for example, it will already trigger the change?

fda-odoo: It is right for files as files will go in the process_queue and then indeed a save event will put it again in it, so processing it later. But for functions, if we are trying to validate them, it means either some code is trying to force a validation manually or that we are actually validating the file. If I'm not wrong I think we can then trigger the arch and arch eval at this moment. In both case we can try to reload the function. If we don't do it, it will never be done as the function is not in the queue (it is in the queue, but via its file symbol, so it will be reloaded when the file is at its validation step, so here)

mmahrouss: Okay yes makes sense, I updated it now

fda-odoo (approved):

PR #179, opened 2025-01-24
[IMP] server: definition for model names

mmahrouss:


server/src/features/definition.rs
```diff
@@ -68,4 +72,30 @@ impl DefinitionFeature {
...
+
+    fn check_for_model_string(session: &mut SessionInfo, eval: &Evaluation, file_symbol: &Rc<RefCell<Symbol>>, links: &mut Vec<Location>) -> bool {
+        let mut model_found = false;
+        if let Some(eval_value) = eval.value.as_ref() {
+            if let crate::core::evaluation::EvaluationValue::CONSTANT(ruff_python_ast::Expr::StringLiteral(expr)) = eval_value {
+                let str = expr.value.to_string();
+                let model = session.sync_odoo.models.get(&str).cloned();
+                if let Some(model) = model {
+                    let from_module = file_symbol.borrow().find_module();
+                    let main_class = model.borrow().get_main_symbols(session, from_module.clone(), &mut None);
```
fda-odoo, line 84: should we go to the main symbol only, or return all symbols that are in dependencies?
I should do the second per default, maybe the first in option imo

fda-odoo (changes requested):

server/src/features/definition.rs
mmahrouss: Updated:)

fda-odoo (approved):

PR #188, opened 2025-01-30
[FIX] client: Default value for disablePythonLanguageServer

mmahrouss:


vscode/client/common/validation.ts
```diff
@@ -12,5 +12,6 @@ export function getConfigurationStructure(id: number = 0) {
         "odooPath": "",
         "addons": [],
         "validatedAddonsPaths": [],
+        "disablePythonLanguageServer": true
```
fda-odoo, line 15: disablePythonLanguageServerPopup or disablePythonLanguageServerNotification would be better no? I feel the naming weird here. And it should be false per default, and true when we don't want the popup anymore

fda-odoo (changes requested):

fda-odoo (approved):

PR #193, opened 2025-02-01
[FIX] server: improve messages managements to not block threads and b…

fda-odoo:
…e able to close

Shutdown event was processed by handle_shutdown from lsp_server. However this method doesn't handle response that could come between the answer of the shutdown request and the exit notification. This commit reimplement the shutdown request processing to allow that. Moreover, send_request method was using specific channel to read response, and so generic channel which contains the exit notification was never read and so the threads could never close. Exit notification is now sent in generic and specific channels for all threads


fda-odoo:
@mmahrouss Now the server should always close when we ask it to do it :)

mmahrouss (approved):

PR #194, opened 2025-02-03
[FIX] server, vscode: Fix server restart on changing python path

mmahrouss:


vscode/client/common/utils.ts
```diff
@@ -161,9 +161,10 @@ export async function buildFinalPythonPath(context, config_python_path: string,
 	}
 
 	//trying to use the VScode python extension
-	if (interpreter && global.IS_PYTHON_EXTENSION_READY !== false) {
+	if (interpreter) {
 		pythonPath = interpreter.path[0]
-		await initializePython(context.subscriptions);
+		await initializePython(context.subscriptions, global.PYTHON_EXTENSION_LISTENER_INSTALLED !== true);
+		global.PYTHON_EXTENSION_LISTENER_INSTALLED = true;
```
fda-odoo, line 167: you never set this to false, so I guess you do not handle uninstallation?

mmahrouss: the flag is to just see if it was installed before or not so I do not install it again. 

So supposedly if it is uninstalled and then installed I do not want to install it again. 

I do not handle installation in the sense that I do not remove the event listener, it is harmless anyway.

but while debugging I noticed it was added more than once, which means everytime it gets triggered one more time

fda-odoo: But then if at some point python extension disappear, we do not fallback on the configuration? That's not a big issue as I guess the actual path stay valid however

mmahrouss: we still fallback on the config, this flag does not affect that. 

the interpreter will be undefined and we will fallback to the else branch, and get the path from the config

fda-odoo (approved):

PR #206, opened 2025-02-06
[FIX] server: unique results in model name hover

mmahrouss:
Also, avoid showing the base class in inherited

### DIFF
--- a/server/src/features/features_utils.rs
+++ b/server/src/features/features_utils.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use ruff_python_ast::{Expr, ExprCall};
 use ruff_text_size::{Ranged, TextRange, TextSize};
 use crate::core::file_mgr::FileMgr;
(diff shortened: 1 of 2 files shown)


server/src/features/features_utils.rs
```diff
@@ -170,31 +170,37 @@ impl FeaturesUtils {
...
                                     value = value + "  \n***  \n" + main_class.doc_string().as_ref().unwrap();
                                 }
-                                let mut other_imps = model.borrow().all_symbols(session, from_module.clone());
+                                let mut other_imps= model.borrow().all_symbols(session, from_module.clone()).into_iter().filter_map(|(sym, needed_module)| {
+                                    if Rc::ptr_eq(&sym, main_class_rc) {
+                                        None // Skip main_class
+                                    } else {
+                                        Some((sym.borrow().find_module().unwrap().borrow().name().clone(), needed_module))
+                                    }
+                                }).collect::<HashSet<_>>().into_iter().collect::<Vec<_>>();
```
mmahrouss, line 186: I can replace collecting to hashset then to vector just to get unique elements by using [itertools::unique](https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.unique) .   
How do you feel about adding a new dependency ?

mmahrouss: Reminder-to-self to use itertools since it is added in a sepearate PR

fda-odoo: I did not add it in my pr, so you can do it here :)

mmahrouss: Oki, will do it and ping you here once it is ready for review

mmahrouss: @fda-odoo Updated

PR #215, opened 2025-02-10
[IMP] server: add logs to understand UnexpectedChar on Uri::from_str(…

fda-odoo:
…uri)


mmahrouss:
I think #216 should already fix the issue, do we still need this?

fda-odoo:
> I think #216 should already fix the issue, do we still need this?

I guess it's always better to have a log than nothing if it appears that it is not fixed?

fda-odoo:
But indeed let's do that in your pr

PR #233, opened 2025-02-12
Named expressions / Walrus Op

mmahrouss:


mmahrouss:
This will conflict wit #249

fda-odoo (approved):

PR #249, opened 2025-03-07
[IMP] server: infer names in outer scope later in file

mmahrouss:


fda-odoo (changes requested):
First commit is nearly deleted by the second one. Maybe move the first (sub_symbols function) as a second commit, based on ast_indexes update?

mmahrouss:
> First commit is nearly deleted by the second one. Maybe move the first (sub_symbols function) as a second commit, based on ast_indexes update?

Do you think that keeping it in just one commit like now is enough though? is it worth it to have a commit just for `visit_sub_stmts`? it feels like a reasonable refactoring to do while moving ast_indexes around, let me know what you think:)

fda-odoo:
Yes, that's ok for me :)

mmahrouss:
> Yes, that's ok for me :)

Okay, I already did it!

fda-odoo (approved):

PR #254, opened 2025-03-13
[FIX] server: fix get_all_inferred_names to get correct completion

mmahrouss:
Handle closures recursively, so not just local scopes and global scope, but traverse out of functions until you hit a file (global scope) Also, fixes duplicated completion on global scope due to local and global scope being identical

### DIFF
--- a/server/src/core/symbols/symbol.rs
+++ b/server/src/core/symbols/symbol.rs
@@ -2072,27 +2072,23 @@ impl Symbol {
     /*
     Return all the symbols that are available at a given position or in a scope for a given start name
      */
-    pub fn get_all_infered_names(odoo: &mut SyncOdoo, on_symbol: &Rc<RefCell<Symbol>>, name: &String, position: Option<u32>) -> Vec<Rc<RefCell<Symbol>>> {
-        let mut results = vec![];
-        //get local symbols
-        on_symbol.borrow().all_symbols().for_each(|sym| {
-            if sym.borrow().name().starts_with(name) {
-                if position.is_none() || !sym.borrow().has_range() || position.unwrap() > sym.borrow().range().end().to_u32() {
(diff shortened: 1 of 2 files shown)


server/src/core/symbols/symbol.rs
```diff
@@ -2035,27 +2035,33 @@ impl Symbol {
...
+        let mut results: Vec<Rc<RefCell<Symbol>>> = vec![];
+        fn helper(on_symbol: &Rc<RefCell<Symbol>>, name: &String, position: Option<u32>, results: &mut Vec<Rc<RefCell<Symbol>>>) {
+            let add_symbols = |results: &mut Vec<Rc<RefCell<Symbol>>>| on_symbol.borrow().all_symbols().for_each(|sym| {
                 if sym.borrow().name().starts_with(name) {
                     if position.is_none() || !sym.borrow().has_range() || position.unwrap() > sym.borrow().range().end().to_u32() {
                         results.push(sym.clone());
                     }
                 }
             });
+            let step_deeper = |results: &mut Vec<Rc<RefCell<Symbol>>>| {
```
fda-odoo, line 2048: I don't really like the "deeper" part of the name, as you are going higher in the tree.
Moreover that function (helper) is a nice recursivity hell. I think you should always try to avoid recursivity when you can. It makes the code harder to read and can hide bugs more easily

mmahrouss: In a way it is recursing the same way as `infer_name`
I will try to make it more readable though. 

I am using the closures to avoid code repetition 

and I have `helper` so that I do not expose the accumulator to the outside signature.

But I will do it another way

mmahrouss: Okay I rewrote in what I think is a more readable way, let me know if it is okay

PR #273, opened 2025-04-03
WIP: Move odoo step 

mmahrouss:


fda-odoo:
The branch randomly doesn't work for models declared in Base
I didn't finish my analysis at this point, but I guess that some models are loaded BEFORE models.py, so the inheritance check doesn't work at arch_eval step (which worked before when done at odoo step)

fda-odoo:
closed by error

mmahrouss:
@fda-odoo 
Maybe you would rename the first commit from draft to something descriptive ?

server/src/core/symbols/file_symbol.rs
```diff
@@ -103,11 +103,13 @@ impl FileSymbol {
...
                 ],
                 vec![ //ARCH_EVAL
-                    None //ARCH
+                    None, //ARCH,
+                    None, //ARCH_EVAL
                 ],
                 vec![
                     None, // ARCH
                     None, //ARCH_EVAL
+                    None, //VALIDATIOn
```
mmahrouss, line 112: small comment typo

PR #275, opened 2025-04-04
[IMP] server: update ruff python parser to 0.11.4

fda-odoo:


fda-odoo:
Note: the new version of Ruff Python Parser include an option "python_version" when parsing file.
It seems it is because some syntax is not retrocompatible. To investiguate

Issue #276, opened 2025-04-07
Recognize Duplicate Manifest Keys


LuvForAirplanes:
I just spent ~20 minutes trying to figure out why my module wasn't initializing correct, then I realized that I had two _depends_ properties in the manifest file, which was seriously messing things up. This really should be checked by Odoo LS.

fda-odoo:
Thank you for the feedback ! That's indeed a good diagnostic we can add ! We'll do it for a next update!

LuvForAirplanes:
Sweet, man! I appreciate it!

<small>This kind of service is unheard of these days. 🤯

PR #277, opened 2025-04-08
[IMP] server: Detect duplicated manifest keys

mmahrouss:
Addressing #276

### DIFF
--- a/server/src/core/symbols/module_symbol.rs
+++ b/server/src/core/symbols/module_symbol.rs
@@ -176,6 +176,7 @@ impl ModuleSymbol {
             ), &session.current_noqa);
             return res;
         }
+        let mut visited_keys = HashSet::new();
         let dict = &ast[0].as_expr_stmt().unwrap().value.clone().dict_expr().unwrap();
         for (index, key) in dict.iter_keys().enumerate() {
             match key {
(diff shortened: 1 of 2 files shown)


fda-odoo:
solves https://github.com/odoo/odoo-ls/issues/276

server/src/core/symbols/module_symbol.rs
```diff
@@ -184,23 +185,27 @@ impl ModuleSymbol {
                     match key {
                         Expr::StringLiteral(key_literal) => {
                             let key_str = key_literal.value.to_string();
+                            if visited_keys.contains(&key_str){
+                                add_diagnostic(&mut res, self._create_diagnostic_for_manifest_key("A manifest should not have duplicate keys", S!("OLS30202"), &key_literal.range, Some(DiagnosticSeverity::WARNING)), &session.current_noqa);
```
fda-odoo, line 189: new code means you should add them to the error code list

fda-odoo (changes requested):

server/src/core/symbols/module_symbol.rs
mmahrouss: Oh I did it but I did not stage it while committing, sorry

mmahrouss: amended, I also had noticed more missing error codes and added them too

fda-odoo (approved):

PR #280, opened 2025-04-11
[IMP] server: Raise error diagnostic on modifying fields in compute

mmahrouss:
methods that are set to modify other fields

### DIFF
--- a/server/src/core/model.rs
+++ b/server/src/core/model.rs
@@ -1,4 +1,5 @@
 use std::cell::RefCell;
+use std::collections::HashMap;
 use std::collections::VecDeque;
 use std::rc::Rc;
 use std::rc::Weak;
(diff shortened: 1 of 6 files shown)


fda-odoo (changes requested):
LGTM.
I don't really like the odoo related code in arch_eval file, but I don't really know how we could move it elsewhere for now.
Maybe you could add a comment there to indicate what this code is doing?

mmahrouss:
> LGTM. I don't really like the odoo related code in arch_eval file, but I don't really know how we could move it elsewhere for now. Maybe you could add a comment there to indicate what this code is doing?

Indeed, but it is for validation for functions so I do not know where else to put it. 

I updated the PR with more comments

fda-odoo (approved):

PR #296, opened 2025-04-23
[FIX] server: fix tree dropping manifest dunder

mmahrouss:


fda-odoo (changes requested):
remove drop_dunder parameter as it should be always true. Fix where you put false

fda-odoo:
@mmahrouss updated, can you review it?

mmahrouss (commented):
LGTM:)

fda-odoo:
@mmahrouss I fixed a deletion, create, rename issues. But I think we should continue to play with it to be sure it's working

fda-odoo (approved):

PR #303, opened 2025-04-25
[FIX] server: find_module should only return modules + improve diagnostic

mmahrouss:


server/src/core/import_resolver.rs
```diff
@@ -185,23 +185,16 @@ pub fn find_module(session: &mut SessionInfo, odoo_addons: Rc<RefCell<Symbol>>,
...
-                }
-            }
+        if !is_dir_cs(full_path.sanitize()) {
+            continue;
+        }
+        let Some(module_symbol) = Symbol::create_from_path(session, &full_path, odoo_addons.clone(), true) else {
+            continue;
+        };
+        if matches!(module_symbol.borrow().typ(), SymType::PACKAGE(PackageType::MODULE)) {
+            session.sync_odoo.modules.insert(name.clone(), Rc::downgrade(&module_symbol));
```
fda-odoo, line 195: I don't think this addition to sync_doo.modules is necessary as create_from_path is already doing it?

mmahrouss: it is redundant indeed

mmahrouss: removed

PR #309, opened 2025-04-29
[FIX] server: fix error on hover with no main EP

mmahrouss:


server/src/core/odoo.rs
```diff
@@ -800,14 +800,16 @@ impl SyncOdoo {
      */
     pub fn get_symbol_of_opened_file(session: &mut SessionInfo, path: &PathBuf) -> Option<Rc<RefCell<Symbol>>> {
         let path_in_tree = path.to_tree_path();
-        for entry in session.sync_odoo.entry_point_mgr.borrow().iter_for_import(session.sync_odoo.entry_point_mgr.borrow().main_entry_point.as_ref().unwrap()) {
-            if (entry.borrow().typ == EntryPointType::MAIN || entry.borrow().addon_to_odoo_path.is_some()) && entry.borrow().is_valid_for(path) {
-                let tree = entry.borrow().get_tree_for_entry(path);
-                let path_symbol = entry.borrow().root.borrow().get_symbol(&tree, u32::MAX);
-                if path_symbol.is_empty() {
-                    continue;
+        if let  Some(main_ep) = session.sync_odoo.entry_point_mgr.borrow().main_entry_point.as_ref(){
```
fda-odoo, line 803: but this way it means that the code won't work anymore for custom entrypoints if main entry is not set

mmahrouss: the if let is only around the first loop actually. 

I am reading carefully. 

So first loop actually only affects MAIN | (some) ADDON Entry points due to the condition

Second loop loops over custom entry points only 

I think I can write the fix to be more readable in the future, by using `iter_main` in the first loop

mmahrouss: updated

server/src/core/odoo.rs
```diff
@@ -800,7 +800,7 @@ impl SyncOdoo {
      */
     pub fn get_symbol_of_opened_file(session: &mut SessionInfo, path: &PathBuf) -> Option<Rc<RefCell<Symbol>>> {
         let path_in_tree = path.to_tree_path();
-        for entry in session.sync_odoo.entry_point_mgr.borrow().iter_for_import(session.sync_odoo.entry_point_mgr.borrow().main_entry_point.as_ref().unwrap()) {
+        for entry in session.sync_odoo.entry_point_mgr.borrow().iter_main() {
             if (entry.borrow().typ == EntryPointType::MAIN || entry.borrow().addon_to_odoo_path.is_some()) && entry.borrow().is_valid_for(path) {
```
mmahrouss, line 804: Due to this `(entry.borrow().typ == EntryPointType::MAIN || entry.borrow().addon_to_odoo_path.is_some())` , we loop over public /builtin eps but always skip them. 

Also since iter_main chains the iterators, it will be just empty if the option is None, so it guards against unwrap errors and also sorts for performance

PR #326, opened 2025-06-30
Tests 0.8.0 mdms

mmahrouss:


server/tests/test_get_symbol.rs
```diff
@@ -76,57 +100,178 @@ fn test_hover_on_model_field_and_method() {
...
+    // Use get_file_info().symbol instead of get_file_symbol
+    let Some(m2_tf_file_symbol) = SyncOdoo::get_symbol_of_opened_file(
+        &mut session,
+        &PathBuf::from(&module2_test_file)
+    ) else {
+        panic!("Failed to get file symbol");
+    };
+
+    // Helper to get hover markdown string at a given (line, character)
+    let get_definition_locs = |session: &mut SessionInfo, f_sym: &Rc<RefCell<Symbol>>, f_info: &Rc<RefCell<FileInfo>>, line: u32, character: u32| {
```
fda-odoo, line 208: I think get_definition_locs and get_hover should maybe be public to all tests.
Maybe create a test_utils.rs file?

fda-odoo (changes requested):

server/tests/test_get_symbol.rs
mmahrouss: Okay I will make it more generic and not a closure and update it

PR #359, opened 2025-07-30
[FIX] multiple files and extends

kdmwangi:
Fix test_config_file_sources_multiple_files_and_extends test where a configuration is extended by multiple children

### DIFF
--- a/server/src/core/config.rs
+++ b/server/src/core/config.rs
@@ -1014,10 +1014,10 @@ fn merge_all_workspaces(
 
             match raw_entry.extends.clone() {
                 Some(v)=>{
-                    merged_raw_config.values_mut().for_each(|ra|{update_python_path(&mut raw_entry,ra, key.clone())});
+                    merged_raw_config.values_mut().for_each(|existing_config|{update_python_path(&mut raw_entry, existing_config, key.clone())});
                 },
                 None => {
-                    merged_raw_config.values_mut().for_each(|ra|{update_python_path(ra,&mut raw_entry, key.clone())});
+                    merged_raw_config.values_mut().for_each(|existing_config|{update_python_path(existing_config, &mut raw_entry, key.clone())});
                 }
             }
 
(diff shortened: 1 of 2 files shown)


server/src/core/config.rs
```diff
@@ -1010,7 +1010,17 @@ fn merge_all_workspaces(
...
+
+            match raw_entry.extends.clone() {
+                Some(v)=>{
+                    merged_raw_config.values_mut().for_each(|ra|{update_python_path(&mut raw_entry,ra, key.clone())});
+                },
+                None => {
+                    merged_raw_config.values_mut().for_each(|ra|{update_python_path(ra,&mut raw_entry, key.clone())});
+                }
+            }
+
```
mmahrouss, line 1023: I am not sure I 100% follow what you are attempting to do. 

As far as I can understand, you are trying to apply the extends behavior across workspace folders? 

Currently, in our design, extends is essentially just being used for the path hierarchy, so vertically, but not horizontally between workspaces. 

If we were to add it accross workspaces, then we would need a mechanism that is more complete such that it affects all parameters and not just python_path.

mmahrouss:
> The basic premise of what I am actually trying to achieve is I took the test driven development approach to try and contribute . So I came across the test_config_file_sources_multiple_files_and_extends which was failing and going through it made absolute sense. Where a child configuration can be defined severally one with extends to the parent and the other without (the both have the same namespace). So the test ensures that the child configuration bothe utilize the same python path from the parent.

Okay I get your point of view. But your changes here are extending python_path accross workspace folders, however, currently we do not do that, extends is exclusively for the file hierarchy starting from root to workspace folder. We do not apply extends among workspaces because the workspace order is not guaranteed. 

One reason the test is failing is correct though, the default python_path being python3 is incorrect. If you wish to contribute by fixing it, I would suggest you make a static lazy value that tries to see if `python3` is valid and falls back to `python`. 

But currently, we do not think the change in `config.rs` is correct. As python_path is already being extended in `apply_extends` which calls `apply_merge`. However, if there is an invalid python_path we fall back to the default.

Thank you so much for your time, we really appreciate it:) looking forward to your reply

PR #362, opened 2025-08-07
[IMP] server: request restart, give shutdown priority

mmahrouss:


server/src/threads.rs
```diff
@@ -319,109 +327,177 @@ pub fn delayed_changes_process_thread(sender_session: Sender<Message>, receiver_
...
-            error!("Got an RecvError, exiting thread");
-            break;
+        // Drain all available messages into buffer
+        loop {
+            let maybe_msg = generic_receiver.try_recv();
+            match maybe_msg {
+                Ok(msg) => {
+                    // Check for shutdown
+                    if matches!(&msg, Message::Notification(n) if n.method.as_str() == Shutdown::METHOD) {
+                        warn!("Main thread - got shutdown.");
```
fda-odoo, line 429: this is not main thread but read thread

server/src/threads.rs
```diff
@@ -319,109 +327,177 @@ pub fn delayed_changes_process_thread(sender_session: Sender<Message>, receiver_
...
-                match r.method.as_str() {
-                    Shutdown::METHOD => { warn!("Read thread - got shutdown."); break;}
-                    _ => {error!("Notification not handled by read thread: {}", r.method)}
+        // If buffer is empty, block for next message so we do not busy wait
+        if buffer.is_empty() {
+            match generic_receiver.recv() {
+                Ok(msg) => {
+                    // Check for shutdown
+                    if matches!(&msg, Message::Notification(n) if n.method.as_str() == Shutdown::METHOD) {
+                        warn!("Main thread - got shutdown.");
```
fda-odoo, line 447: same

PR #371, opened 2025-08-11
[IMP] server: refactor ast_indexes into new node_index support from Ruff

fda-odoo:
Drop our home-made ast_indexes that help us indexing statements and use built-in node_indexes that are now directly on Nodes in last version of Ruff_python_ast.

### DIFF
--- a/server/src/core/symbols/function_symbol.rs
+++ b/server/src/core/symbols/function_symbol.rs
@@ -1,7 +1,7 @@
 use std::{cell::RefCell, cmp::min, collections::HashMap, rc::{Rc, Weak}};
 
 use lsp_types::Diagnostic;
-use ruff_python_ast::{Expr, ExprCall};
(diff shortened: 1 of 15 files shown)


server/src/core/symbols/function_symbol.rs
```diff
@@ -34,7 +34,7 @@ pub struct FunctionSymbol {
     pub is_static: bool,
     pub is_property: bool,
     pub doc_string: Option<String>,
-    pub ast_indexes: Vec<u16>, //list of index to reach the corresponding ast node from file ast
+    pub ast_indexes: AtomicNodeIndex,
```
mmahrouss, line 37: Suggestion: Maybe we could rename it to node_index? since it is just one index, not a sequence of indexes? also at symbol, class symbol. variable symbol ... ?

PR #376, opened 2025-08-12
[IMP] server: use current_exe instead of current_dir in config

fda-odoo:


fda-odoo:
not ready yet, as does not work for tests and debug mode

fda-odoo:
@mmahrouss Let me know what you think about this PR
We should use search for stdlib and stubs next to the executable instead of inside the current user directory, but as it's breaking debug and tests, I put the previous behaviour as a fallback

mmahrouss (approved):
It does look good. and it should be more robust/flexible as it searches next to the current executable then checks the CWD.

PR #399, opened 2025-09-03
[IMP] remove read thread

mmahrouss:


server/src/server.rs
```diff
@@ -421,9 +383,9 @@ impl Server {
                     },
                     DocumentSymbolRequest::METHOD => {
                         if DEBUG_THREADS {
-                            info!("Sending request to read thread : {} - {}", r.method, r.id);
+                            info!("Sending request to main thread : {} - {}", r.method, r.id);
```
fda-odoo, line 386: it can be merged with previous match case then

PR #416, opened 2025-09-12
[FIX] do not attempt to create custom EP on a directory while renaming

mmahrouss:


mmahrouss:
Note-to-self: 

on a folder far away from odoo, create `__manifest__.py` then `__init__.py`. then on renaming or deletion of `__init__.py` the server panics. comes from `add_from_self_reload` not expecting NAMESPACE

server/src/core/entry_point.rs
```diff
@@ -148,8 +148,10 @@ impl EntryPointMgr {
         EntryPointMgr::create_dir_symbols_from_path_to_entry(session, &PathBuf::from(path), entry)
     }
 
-    pub fn create_new_custom_entry_for_path(session: &mut SessionInfo, path: &String) -> bool {
-        let path_sanitized = PathBuf::from(path).sanitize();
+    /// Create a new custom entry point for a given tree path and file path.
+    /// tree_path can possibly be the path stripped from __manifest__/__init__.py
+    pub fn create_new_custom_entry_for_path(session: &mut SessionInfo, tree_path: &String, file_path: &String) -> bool {
+        let path_sanitized = PathBuf::from(tree_path).sanitize();
```
fda-odoo: Why do you sanitize it again if all call to the function sanitize it already? Usually each function assume that path is sanitized or not

fda-odoo (approved):

server/src/core/entry_point.rs
mmahrouss: You are right! Will update it

mmahrouss: it was already there, but it is indeed unnecessary

PR #433, opened 2025-10-01
[FIX] adapt for file versions starting at 0

mmahrouss:


server/src/core/file_mgr.rs
```diff
@@ -114,15 +114,17 @@ impl FileInfo {
         // Return true if the update has been done and not discarded
         if let Some(version) = version {
             if version == -100 {
-                self.version = 1;
+                self.version = Some(1);
```
fda-odoo: but then are we sure that it should be 1 here? what if we receive an update with 0 by after?

mmahrouss: good question, I have to consider that, let me think

mmahrouss: I think yes, it will be safe to keep it a negative number

server/src/core/file_mgr.rs
```diff
@@ -114,15 +114,17 @@ impl FileInfo {
         // Return true if the update has been done and not discarded
         if let Some(version) = version {
             if version == -100 {
```
fda-odoo: maybe we should prevent updating version to -100 if file is currently opened, no?

mmahrouss: But how can that happen, isn't that a violation of the protocol?

fda-odoo: a version of -100 is set by ourself, not by the IDE, usually when we are manually opening it.

mmahrouss: ah, okay, then indeed it would be better to have the guard, I will add it

PR #435, opened 2025-10-01
[IMP] Autocompletion on import statements

fda-odoo:


server/Cargo.toml
```diff
@@ -53,5 +53,5 @@ iai-callgrind = "0.14.2"
 assert_fs = "1.1"
 
 [features]
-default = []
+default = ["debug_yarn"]
```
mmahrouss: Did you push this intentionally?

server/src/constants.rs
```diff
@@ -6,11 +6,11 @@ pub const EXTENSION_NAME: &str = "Odoo";
...
 
 pub const DEBUG_ODOO_BUILDER: bool = false;
-pub const DEBUG_MEMORY: bool = false;
+pub const DEBUG_MEMORY: bool = true;
 pub const DEBUG_THREADS: bool = false;
 pub const DEBUG_STEPS: bool = false;
 pub const DEBUG_STEPS_ONLY_INTERNAL: bool = true;
-pub const DEBUG_REBUILD_NOW: bool = false;
+pub const DEBUG_REBUILD_NOW: bool = true;
 pub const DEBUG_BORROW_GUARDS: bool = false;
```
mmahrouss: Same with those? do you want to have them enabled on Alpha? and we'll revert them on release?

server/src/core/import_resolver.rs
```diff
@@ -434,64 +442,189 @@ fn _resolve_new_symbol(session: &mut SessionInfo, parent: Rc<RefCell<Symbol>>, n
...
+                if split.len() > 1 {
+                    (Some(Identifier::new(split[0..split.len()-1].join(".").as_str(), TextRange::default())), split.last().unwrap().to_string())
+                } else {
+                    (None, split.last().unwrap().to_string())
+                }
+            } else {
+                (Some(Identifier::new(from_stmt_inner.clone(), TextRange::default())), import.clone())
+            }
+        },
+        None => (None, S!("")),
```
mmahrouss: Shouldn't this be `import.clone()` ? 

Like that `import odo|` will try to list everything because it is starting from empty string? 
```suggestion
        None => (None, import.clone()),
```

server/Cargo.toml
fda-odoo: :sweat_smile:

server/src/constants.rs
fda-odoo: :sweat_smile: :sweat_smile:

mmahrouss (approved):

fda-odoo:
closed for https://github.com/odoo/odoo-ls/pull/443

PR #436, opened 2025-10-03
[IMP] Filter out magic fields from go to definition

mmahrouss:
Task: 5111939

### DIFF
--- a/server/src/core/python_odoo_builder.rs
+++ b/server/src/core/python_odoo_builder.rs
@@ -17,6 +17,15 @@ use crate::{oyarn, Sy, S};
 
 use super::evaluation::{ContextValue, Evaluation, EvaluationSymbolPtr, EvaluationValue};
 
+pub const MAGIC_FIELDS: [&str; 6] = [
+    "id",
+    "display_name",
+    "create_uid",
+    "create_date",
+    "write_uid",
+    "write_date"
+];
+
 pub struct PythonOdooBuilder {
     symbol: Rc<RefCell<Symbol>>,
 }
--- a/server/src/features/definition.rs
+++ b/server/src/features/definition.rs
@@ -5,7 +5,7 @@ use std::path::PathBuf;
 use std::{cell::RefCell, rc::Rc};
 
 use crate::constants::SymType;
-use crate::core::evaluation::{Evaluation, EvaluationValue};
+use crate::core::evaluation::{Evaluation, EvaluationValue, ExprOrIdent};
 use crate::core::file_mgr::{FileInfo, FileMgr};
 use crate::core::odoo::SyncOdoo;
 use crate::core::python_odoo_builder::MAGIC_FIELDS;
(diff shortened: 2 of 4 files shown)


server/src/core/python_odoo_builder.rs
```diff
@@ -17,6 +17,15 @@ use crate::{oyarn, Sy, S};
 
 use super::evaluation::{ContextValue, Evaluation, EvaluationSymbolPtr, EvaluationValue};
 
+pub const MAGIC_FIELDS: [&str; 6] = [
+    "id",
+    "display_name",
```
andg-odoo, line 22: How possible would it be for display_name to go to any `_compute_display_name` methods associated with the model?

mmahrouss: @andg-odoo, Well, I think it is possible, it is a bit of a custom behavior though, but it has some sense to it. I shall discuss with @fda-odoo and see what we think about implementing that.  
Thanks for the suggestion

fda-odoo: That's indeed a good idea, I added it to our todolist :)

server/src/core/evaluation.rs
```diff
@@ -575,7 +575,7 @@ impl Evaluation {
             (S!("module"), from_module),
             (S!("range"), ContextValue::RANGE(ast.range()))
         ]));
-        let analyze_result = Evaluation::analyze_ast(session, &ExprOrIdent::Expr(ast), parent, max_infer, &mut context, for_annotation, required_dependencies);
+        let analyze_result = Evaluation::analyze_ast(session, &ExprOrIdent::Expr(ast.clone()), parent, max_infer, &mut context, for_annotation, required_dependencies);
```
fda-odoo: Why do you need cloning ast everywhere? That's an expensive change

mmahrouss: Too much moving of reference was a bit scary. I am trying to refactor it now safely

server/src/features/definition.rs
```diff
@@ -148,19 +149,90 @@ impl DefinitionFeature {
...
+        let crate::core::evaluation::ExprOrIdent::Expr(Expr::Attribute(attr_expr)) = expr else {
+            return;
+        };
+        let (analyse_ast_result, _range) = AstUtils::get_symbols(session, file_symbol, offset as u32, &crate::core::evaluation::ExprOrIdent::Expr(&attr_expr.value));
+        let eval_ptrs = analyse_ast_result.evaluations.iter().flat_map(|eval| Symbol::follow_ref(eval.symbol.get_symbol_ptr(), session, &mut None, false, false, None)).collect::<Vec<_>>();
+        let symbols = eval_ptrs.iter().flat_map(|eval_ptr| {
+            let Some(symbol) = eval_ptr.upgrade_weak() else {
+                return  vec![];
+            };
+            symbol.borrow().get_member_symbol(session, &S!("_compute_display_name"), None/*TODO */, false, false, true, false).0
```
fda-odoo: TODO?

fda-odoo (approved):

PR #439, opened 2025-10-07
[IMP] report recent messages on crash report

mmahrouss:


server/src/server.rs
```diff
@@ -293,6 +293,13 @@ impl Server {
             }
 
             if index == 0 { //comes from client
+                // Save Requests and Notifications to crash buffer
+                match &msg {
+                    Message::Request(_) | Message::Notification(_) => {
+                        crash_buffer::push_message(msg.clone());
+                    },
+                    _ => {}
```
fda-odoo: Why omit Response? They could be useful too

mmahrouss: Indeed, I will push everything

PR #440, opened 2025-10-13
[IMP] handle untitled python files

mmahrouss:


server/src/core/entry_point.rs
```diff
@@ -359,6 +406,9 @@ impl EntryPoint {
     }
 
     pub fn is_valid_for(&self, path: &PathBuf) -> bool {
+        if self.typ == EntryPointType::UNTITLED {
```
fda-odoo, line 403: why is_valid_for is not usable for untitled entry point?

fda-odoo (changes requested):
There is a lot of duplicated code, or at least redundant ones.
Why double most of the route as "update_file_cache" and "update_untitled_file_cache"?
I would expect the actual update_file_cache to handle untitled files.
Maybe we should include scheme in path? Moreover it would let space for new schemes in the future

server/src/core/entry_point.rs
mmahrouss: updated

server/src/core/entry_point.rs
```diff
@@ -25,8 +26,27 @@ impl EntryPointMgr {
             main_entry_point: None,
             addons_entry_points: vec![],
             custom_entry_points: vec![],
+            untitled_entry_points: vec![],
         }
     }
+    /// Create a new entry for an untitled (in-memory) file.
+    /// Returns the file symbol for the untitled entry.
+    pub fn add_entry_to_untitled(session: &mut SessionInfo, path: String) -> Rc<RefCell<Symbol>> {
+        // For untitled files, we use a minimal tree: just the name as a single OYarn
```
fda-odoo, line 35: info!("Adding new untitled entry point: {}", path);

fda-odoo (approved):

PR #442, opened 2025-10-14
[IMP] model not found and validation of csv records

mmahrouss:


fda-odoo (approved):

server/src/core/symbols/module_symbol.rs
```diff
@@ -352,6 +354,38 @@ impl ModuleSymbol {
         diagnostics
     }
 
+    pub fn validate_manifest(symbol: &Rc<RefCell<Symbol>>, session: &mut SessionInfo){
+        let data_paths = symbol.borrow().as_module_package().data.clone();
```
fda-odoo, line 369: Do you really need a clone here?

fda-odoo (approved):

server/src/core/symbols/module_symbol.rs
mmahrouss: I think we do, because later on we mutably borrow symbol 
`symbol.borrow_mut().as_module_package_mut()`

we also do the same in `check_data` and `load_data` for the same reason

PR #443, opened 2025-10-14
Alpha imports hover fda

fda-odoo:


server/src/features/ast_utils.rs
```diff
@@ -58,6 +64,98 @@ impl AstUtils {
...
+    fn get_symbol_in_import(session: &mut SessionInfo, file_symbol: &Rc<RefCell<Symbol>>, offset: u32, stmt: &Stmt) -> Option<(AnalyzeAstResult, Option<TextRange>, Option<ExprCall>)> {
+        match stmt {
+            //for all imports, the idea will be to check if we are on the last name of the import (then it has benn imported already and we can fallback on it),
+            //or then take the full tree to the offset symbol and resolve_import on it as it was in a 'from' clause.
+            Stmt::Import(stmt) => {
+                for alias in stmt.names.iter() {
+                    if alias.range().contains(TextSize::new(offset)) {
+                        let mut is_last = false;
+                        let (to_analyze, range) = if alias.name.range().contains(TextSize::new(offset)) {
+                            let next_dot_offset = alias.name.id.as_str()[offset as usize - alias.range().start().to_usize()..].find(".");
```
mmahrouss: Shouldn't we get the range of the name, not the whole alias? They probably have the same start, it is probably okay either way. Maybe for readability? 
```rust
alias.name.range().start().to_usize()
```

mmahrouss (approved):

PR #445, opened 2025-10-15
[IMP] hover and definition for module names in manifest

mmahrouss:


server/src/features/definition.rs
```diff
@@ -90,6 +90,33 @@ impl DefinitionFeature {
         model_found
     }
 
+    fn check_for_module_string(session: &mut SessionInfo, eval: &Evaluation, file_symbol: &Rc<RefCell<Symbol>>, links: &mut Vec<LocationLink>) -> bool {
+        if file_symbol.borrow().typ() != SymType::PACKAGE(PackageType::MODULE) {
```
fda-odoo: But this doesn't block the __init__.py of the module?

Issue #418, opened 2025-09-16
GoToDef not supported on `inverse_name`


andg-odoo:
Post OXP feature/bug request (Not 1.0 blocking):

**Your setup**
Version: main branch via `cargo install --git`
Operating System: Ubuntu 24.04
IDE and/or Integration tool (for example: Vscode - official extension): neovim lsp

**Describe the bug**
On field definitions, `comodel_name`, `related`, `compute`, and many other parameters are supported and properly handle hover or hinting. This is missing on `inverse_name` for `One2Many`.

**To Reproduce**
Try to gotodef on `inverse_name` or look at hover.

**Expected behavior**
GoToDef should be able to handle it as we already are able to pull the model.

andg-odoo:
Mainly fixed in #446, requiring #515 for non kwarg `inverse_name`

andg-odoo:
#515 has been merged and now works in 1.2.1

PR #446, opened 2025-10-16
Features for inverse_name

mmahrouss:


mmahrouss:
Related https://github.com/odoo/odoo-ls/issues/418

fda-odoo:
@mmahrouss it doens't compile with yarn ;)

server/src/core/symbols/symbol.rs
```diff
@@ -2779,6 +2800,9 @@ impl Symbol {
             if only_fields {
                 content_syms = content_syms.iter().filter(|x| x.borrow().is_field(session)).cloned().collect();
             }
+            if only_methods {
+                content_syms = content_syms.iter().filter(|x| x.borrow().typ() == SymType::FUNCTION).cloned().collect();
```
fda-odoo: what about a variable that is evaluated to a function definition?

mmahrouss: Updated!

PR #454, opened 2025-10-23
[IMP] use workDoneProgress to report loading progression and git lock…

fda-odoo:
… events


mmahrouss (approved):
LGTM, Might be nice to check client capabilities for it. We do not do it for other features, but we could start here

PR #460, opened 2025-10-29
[IMP] Workspace Symbol request and $/Cancel support

fda-odoo:


server/src/core/file_mgr.rs
```diff
@@ -397,24 +397,50 @@ impl FileInfo {
...
+        let column = char - first_char_of_line;
+        Some(Position::new(line as u32, column as u32))
+    }
+
     pub fn offset_to_position(&self, offset: usize) -> Position {
         FileInfo::offset_to_position_with_rope(self.file_info_ast.borrow().text_rope.as_ref().expect("no rope provided"), offset)
     }
 
+    pub fn try_offset_to_position(&self, offset: usize) -> Option<Position> {
+        FileInfo::try_offset_to_position_with_rope(self.file_info_ast.borrow().text_rope.as_ref().expect("no rope provided"), offset)
```
mmahrouss: Maybe you want to also not expect here? 

Perhaps:
```suggestion
        FileInfo::try_offset_to_position_with_rope(self.file_info_ast.borrow().text_rope.as_ref()?, offset)
```

server/src/features/workspace_symbols.rs
```diff
@@ -0,0 +1,186 @@
...
+                    code: ErrorCode::RequestCanceled as i32,
+                    message: S!("Workspace Symbol request cancelled"),
+                    data: None,
+                });
+            }
+        }
+        Ok(Some(WorkspaceSymbolResponse::Nested(symbols)))
+    }
+
+    fn browse_symbol(session: &mut SessionInfo, symbol: &Rc<RefCell<Symbol>>, query: &String, parent: Option<String>, parent_path: Option<&String>, can_resolve_location_range: bool, results: &mut Vec<WorkspaceSymbol>) -> bool {
```
mmahrouss, line 47: Could be nice to explain what the boolean means? 

It is slightly confusing that it returns true if it is cancelled

PR #478, opened 2025-11-07
[IMP] add debug configurations for rust-gdb

cammarosano:
While still not par with window's cppvsdbg, rust-gdb has some advantages over lldb, specially when inspecting an enum type.

Make sure you have the `ms-vscode.cpptools` extension installed in order to test it.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,6 +29,46 @@
             "console": "externalTerminal"
         },
         {
+            "name": "Launch Server (gdb)",
+            "type": "cppdbg",
+            "request": "launch",
+            "preLaunchTask": "cargo build",
+            "program": "${workspaceFolder}/server/target/debug/odoo_ls_server",
+            "args": ["--use-tcp"],
+            "cwd": "${workspaceFolder}/server",
+            "externalConsole": false,
+            "MIMode": "gdb",
+            "miDebuggerPath": "rust-gdb",
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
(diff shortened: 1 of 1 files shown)


mmahrouss (approved):

.vscode/launch.json
```diff
@@ -28,6 +28,46 @@
...
+            "name": "Launch Server (gdb)",
+            "type": "cppdbg",
+            "request": "launch",
+            "preLaunchTask": "cargo build",
+            "program": "${workspaceFolder}/server/target/debug/odoo_ls_server",
+            "args": ["--use-tcp"],
+            "cwd": "${workspaceFolder}/server",
+            "externalConsole": false,
+            "MIMode": "gdb",
+            "miDebuggerPath": "/usr/bin/rust-gdb",
```
mmahrouss: For me rust-gdb was not installd under `/usr/bin/rust-gdb` but rather under `~/.cargo/bin/rust-gdb`.  
Regardless setting it to `"miDebuggerPath": "rust-gdb",` worked for me as rust-gdb is in `PATH`

cammarosano: I guess it should be on the `PATH`.
I've updated it to `"miDebuggerPath": "rust-gdb"`

.vscode/launch.json
```diff
@@ -45,6 +45,11 @@
                     "text": "-enable-pretty-printing",
                     "ignoreFailures": true
                 },
+                {
+                    "description": "Set print depth limit to prevent infinite recursion",
```
cammarosano, line 49: LLM-generated text/explanation, but I am not fully convinced that's really the case.

PR #484, opened 2025-11-21
List namespace paths on hover + path-related refactors

cammarosano:


server/tests/test_get_symbol.rs
```diff
@@ -150,6 +150,72 @@ fn test_hover_on_model_field_and_method() {
     );
 }
 
+#[test]
+fn test_hover_on_namespace_and_module() {
```
cammarosano, line 191: The reason I placed this test here is the similarities with the test above (test_hover_on_model_field_and_method).
But should we have a test file per feature, both tests would end up in a test_hover file.

server/src/features/features_utils.rs
```diff
@@ -674,24 +673,41 @@ impl FeaturesUtils {
...
     }
 
-    /// Finds and returns useful links for an evaluation
-    fn get_useful_link(_session: &mut SessionInfo, typ: &EvaluationSymbolPtr) -> String {
+    /// Finds and returns useful links or directory locations for an evaluation
+    fn get_location_info(_session: &mut SessionInfo, typ: &EvaluationSymbolPtr) -> String {
         // Possibly add more links in the future
         let Some(typ) = typ.upgrade_weak() else {
-            return S!("")
+            return S!("");
```
fda-odoo: That's not useful :)

server/src/core/import_resolver.rs
```diff
@@ -223,7 +228,7 @@ pub fn resolve_import_stmt(session: &mut SessionInfo, source_file_symbol: &Rc<Re
 }
 
 pub fn find_module(session: &mut SessionInfo, odoo_addons: Rc<RefCell<Symbol>>, name: &OYarn) -> Option<Rc<RefCell<Symbol>>> {
-    let paths = (*odoo_addons).borrow().paths().clone();
+    let paths = (*odoo_addons).borrow().path().as_vec();
```
cammarosano: or `.as_multiple`, if we're sure that `odoo_addons` is a namespace?

server/src/core/odoo.rs
```diff
@@ -468,8 +469,10 @@ impl SyncOdoo {
     fn build_modules(session: &mut SessionInfo) {
         {
             let addons_symbol = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &tree(vec!["odoo", "addons"], vec![]), u32::MAX)[0].clone();
-            let addons_path = addons_symbol.borrow().paths().clone();
-            for addon_path in addons_path.iter() {
+            let SymbolPath::Multiple(addons_paths) = addons_symbol.borrow().path() else {
+                panic!("Odoo addons symbol paths should be a namespace");
+            };
```
cammarosano: @fda-odoo we wrote this together.
But since then `SymbolPath::as_multiple` has been introduced.
Shall I replace this by simply `let addon_paths = addons_symbol.borrow().path().as_multiple()`?

server/src/core/odoo.rs
```diff
@@ -756,7 +759,7 @@ impl SyncOdoo {
 
     pub fn add_to_rebuild_arch(&mut self, symbol: Rc<RefCell<Symbol>>) {
         if DEBUG_THREADS {
-            trace!("ADDED TO ARCH - {}", symbol.borrow().paths().first().unwrap_or(&symbol.borrow().name().to_string()));
+            trace!("ADDED TO ARCH - {} - {:?}", symbol.borrow().name(), symbol.borrow().path());
```
cammarosano: @fda-odoo I had shown you a `Symbol::debug_paths()` just for this cases.
But I decided to drop it , and use this instead (leveraging the derived `debug` trait on `SymbolPath` and always displaying the symbol's name anyway). What do you think?

server/src/core/import_resolver.rs
```diff
@@ -82,7 +85,8 @@ pub fn resolve_from_stmt(session: &mut SessionInfo, source_file_symbol: &Rc<RefC
         level,
         from_stmt);
     drop(_source_file_symbol_lock);
-    let source_path = source_file_symbol.borrow().paths()[0].clone();
+    // source_file_symbol is either a file or a namespace with a single path
+    let source_path = source_file_symbol.borrow().path().as_vec()[0].clone();
```
cammarosano: This simply keeps the old behavior. But it feels dodgy. What do think?
See also: [the previous comment](https://github.com/odoo/odoo-ls/pull/484/files#r2550524957)

server/src/core/import_resolver.rs
```diff
@@ -139,7 +143,8 @@ pub fn resolve_import_stmt(session: &mut SessionInfo, source_file_symbol: &Rc<Re
         } else {
             vec![]
         };
-        let source_path = source_file_symbol.borrow().paths()[0].clone();
+        // source_file_symbol is either a file or a namespace with a single path
+        let source_path = source_file_symbol.borrow().path().as_vec()[0].clone();
```
cammarosano: Same as above (preserves the original behavior). Dodgy?
See also: [manual_import](https://github.com/odoo/odoo-ls/pull/484/files#r2550524957)

server/src/core/symbols/symbol.rs
```diff
@@ -3171,3 +3177,35 @@ impl Symbol {
         res
     }*/
 }
+
+impl SymbolPath {
+    pub fn as_single(self) -> String {
```
fda-odoo: It is not really intuitive that "as_single" returns a String, and not a Single struct.
If we want a String, maybe 
```rust
pub fn as_string(self) {
    self.as_single().map(|s| s.to_owned())
}
```
would be nice to add? or a match in as_string is ok too.

fda-odoo: Or we could rename as_single to get_single_string

server/src/core/import_resolver.rs
```diff
@@ -57,6 +57,9 @@ fn resolve_import_stmt_hook(alias: &Alias, from_symbol: &Option<Rc<RefCell<Symbo
  * Helper to manually import a symbol. Do not forget to use level instead of '.' in the from_stmt parameter.
  */
 pub fn manual_import(session: &mut SessionInfo, source_file_symbol: &Rc<RefCell<Symbol>>, from_stmt:Option<String>, name: &str, asname: Option<String>, level: u32, diagnostics: &mut Option<&mut Vec<Diagnostic>>) -> Vec<ImportResult> {
+    if source_file_symbol.borrow().path().as_vec().len() > 1 {
+        panic!("Importing from a multi-path namespace symbol is not supported in manual_import");
+    }
```
cammarosano: As a reminder: this makes it possible that `source_file_symbol` is a namespace symbol (thus, with Multiple paths) in `resolve_from_stmt` and `resolve_import_stmt`

server/src/core/symbols/symbol.rs
```diff
@@ -3171,3 +3177,35 @@ impl Symbol {
...
+            Self::Single(s) => s,
+            _ => panic!("Not a Single"),
+        }
+    }
+
+    pub fn is_single(&self) -> bool {
+        matches!(self, Self::Single(_))
+    }
+
+    pub fn as_multiple(self) -> Vec<String> {
```
fda-odoo: Same, maybe get_multiple_vector?

fda-odoo (changes requested):
First part review, I didn't review the usages already, only the first 4 commits and the struct SymbolPath.
I'll continue newt week

server/src/core/symbols/symbol.rs
```diff
@@ -3188,3 +3177,35 @@ impl Symbol {
...
+            Self::Multiple(v) => v,
+            _ => panic!("Not a Multiple"),
+        }
+    }
+
+    pub fn is_multiple(&self) -> bool {
+        matches!(self, Self::Multiple(_))
+    }
+
+    pub fn as_vec(self) -> Vec<String> {
```
cammarosano: This enables us to use the result of `.path()` just like the old `paths()` - which is probably a bad thing. :grimacing:

server/src/core/symbols/symbol.rs
```diff
@@ -3188,3 +3177,35 @@ impl Symbol {
...
     }*/
 }
+
+impl SymbolPath {
+    pub fn as_single(self) -> String {
+        match self {
+            Self::Single(s) => s,
+            _ => panic!("Not a Single"),
+        }
+    }
```
cammarosano: This consumes `self` and returns the `String`. Which is usually what we want when chaining with `.path()` (right?).
But then this clashes with Rust's convention of returning references for methods called `as_*`.
Maybe this should be called `to_single` or `unwrap_single` or something like this?
I don't particularly like those names, but it feels closely related to what `Option::unwrap` does.
Alternatively, should this be returning a `&String` (in which case `Symbol::paths` should probably also return `&String`)?

server/src/features/features_utils.rs
cammarosano: oops, sorry, automatic formatting (rust-fmt) :grimacing:

server/src/core/symbols/symbol.rs
cammarosano: According to what we discussed live, iirc you suggested to rename it to "as_single_string", and @mmahrouss  suggested "**to**_single_string", because "as\_" suggests returning a reference.
So the last push has "to_single_string" the as name for this method.
I'm fine with both, but I'd probably name it `unwrap_single`, as it closely relates to Options and Results's `unwrap`, in which you unwrap the Some/Ok value or panic. In our case, because we have more than one alternative to None (Single and Multiple), we'd have to be specific about what we're unwrapping (`unwrap_single` or `unwrap_multiple`).

fda-odoo:
@cammarosano Can you rebase this one too?

cammarosano:
> @cammarosano Can you rebase this one too?

done

fda-odoo (approved):
Well done, that's working well !

PR #488, opened 2025-11-25
Handle multiple import evaluations, add stop_on_tree to follow_ref, Handle env get item for multiple fn evals

mmahrouss:
Closes https://github.com/odoo/odoo-ls/issues/487

PR #497, opened 2025-12-03
[IMP] controllers

cammarosano:
This PR adds some missing features related to controllers.
- `request` now evaluates to `Request` class
- `request.env` now evaluates to `Environment | None`
- subscript on request's env is now supported. E.g. `request.env["res.partner"]` evaluates to the model 

Next steps:
In a Controller, `self.env` still does not work, as `env` is a property (which simply returns `request.env`), and properties are not resolved to their return types.
Support for resolving property types is introduced by https://github.com/odoo/odoo-ls/pull/511, which should then:
- enable `self.env` in a `Controller` class
- enable `request.env` before 15.3, in which `request` is an instance of `WebRequest`, which in turn has `env` as a property.

### DIFF
--- a/server/src/core/python_arch_builder_hooks.rs
+++ b/server/src/core/python_arch_builder_hooks.rs
@@ -38,12 +38,26 @@ static arch_class_hooks: Lazy<Vec<PythonArchClassHook>> = Lazy::new(|| {vec![
                 let mut range = symbol.borrow().range().clone();
                 let slots = symbol.borrow().get_symbol(&(vec![], vec![Sy!("__slots__")]), u32::MAX);
                 if slots.len() == 1 {
-                    if slots.len() == 1 {
-                        range = slots[0].borrow().range().clone();
-                    }
+                    range = slots[0].borrow().range().clone();
                 }
                 symbol.borrow_mut().add_new_variable(session, Sy!("env"), &range);
             }
+        }
+    },
+    PythonArchClassHook {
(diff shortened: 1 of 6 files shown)


cammarosano:
I've just realized that the 2nd commit in this PR handles the same issue as @mmahrouss 's PR: https://github.com/odoo/odoo-ls/pull/488
Dropping it. For the record: https://github.com/odoo/odoo-ls/pull/497/commits/80b805c9b4d8083ef786eb2db21f6aaf783df47d.

cammarosano:
This PR must be adapted to https://github.com/odoo/odoo/pull/240683, that moves `request` (and `Request`) from http.py to requestlib.py
Edit: done.

andg-odoo:
@cammarosano Hi! I've been doing a few PRs recently on the IAP side which is basically all controllers, and definitely missing the environment access on request. Just wondering if this PR is still in the works? I checked out the branch locally and it's been working well for what I need for the past bit so it would be nice to have it merged with all the other cool features of alpha.

Thanks again for all your work on this! :)

fda-odoo:
We'll try to include this branch in the 1.3. @cammarosano can you rebase it on alpha?

cammarosano:
> We'll try to include this branch in the 1.3. @cammarosano can you rebase it on alpha?

Done

fda-odoo (approved):
well done !

PR #498, opened 2025-12-05
[FIX] do not exit on missing python command

mmahrouss:


fda-odoo (approved):
Nice! That was indeed needed for a long time already.
Allowing OdooLS to run without python will lead to some missing imports, and it is maybe not really clear for the user.
I was thinking about adding a notification, but I don't like them.
Maybe we should be sure to at least display that python is invalid in the hover menu of OdooLS?

mmahrouss:
> Nice! That was indeed needed for a long time already. Allowing OdooLS to run without python will lead to some missing imports, and it is maybe not really clear for the user. I was thinking about adding a notification, but I don't like them. Maybe we should be sure to at least display that python is invalid in the hover menu of OdooLS?

Okay, I think it is better to do it in another task, I will create one

PR #500, opened 2025-12-08
[FIX] do not error on SQL data files

mmahrouss:


cammarosano:
Hey! I've just noticed that `load_data` @ module_symbol.rs will log an error for not recognizing anything other than .xml and .csv as extensions for module data.

https://github.com/odoo/odoo-ls/blob/56eb1112b1bbca9cd20eb5991e4bbacd02a82a38/server/src/core/symbols/module_symbol.rs#L441

You might wanna handle that too in your PR.

mmahrouss:
> Hey! I've just noticed that `load_data` @ module_symbol.rs will log an error for not recognizing anything other than .xml and .csv as extensions for module data.
> 
> https://github.com/odoo/odoo-ls/blob/56eb1112b1bbca9cd20eb5991e4bbacd02a82a38/server/src/core/symbols/module_symbol.rs#L441
> 
> You might wanna handle that too in your PR.

Yes, I will have to not error on SQL files, but rather do nothing

fda-odoo (approved):

PR #501, opened 2025-12-08
[IMP] handle model subscripts

mmahrouss:


fda-odoo (approved):

server/tests/test_get_symbol.rs
```diff
@@ -142,8 +142,15 @@ fn test_hover_on_model_field_and_method() {
         "Hover on field_name in search domain should show field name and field type"
     );
 
+    //Hover on self.env with res.partner with subscript
+    let hover_partner = test_utils::get_hover_markdown(&mut session, &file_symbol, &file_info, 34, 10).unwrap_or_default();
```
fda-odoo: I don't think you should use the "hover" feature to test this.
If hover feature changes (even only the way we display it), we'll have to update this test that is not related.
I think you should compare the evaluation of the variable partner directly (you can use AstUtils::get_symbol, but I wouldn't use hover directly)

fda-odoo: moreover compare strings seems a weaker comparison that the classes themselves

mmahrouss: @fda-odoo Okay I updated it, but now this PR depends on https://github.com/odoo/odoo-ls/pull/520 being merged first, I changed the targeted branch

fda-odoo (approved):

PR #507, opened 2025-12-10
[FIX] fix failure to detect models when CachedModel is missing

mmahrouss:
Since we added CachedModel for 19.1. All commit that target 19.1 that were not rebased on master after adding the CachedModel were failing because we were always returning false in test_symbol_is_model because we do not find CachedModel. 

However, I think it is incorrect to return false anyway, when CachedModel does not exist. Because a model can inherit from models.Model, and be fine, even before CachedModel is needed, same with TransientModel.

### DIFF
--- a/server/src/core/python_odoo_builder.rs
+++ b/server/src/core/python_odoo_builder.rs
@@ -398,27 +398,22 @@ impl PythonOdooBuilder {
         let base_model_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &base_model_tree, u32::MAX);
         let model_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &model_tree, u32::MAX);
         let transient_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &transient_tree, u32::MAX);
-        if base_model_syms.is_empty() || model_syms.is_empty() || transient_syms.is_empty() {
-            //one of them is not already loaded, but that's not really an issue, as now odoo step has been merged
-            //with arch eval step, some files will be odooed before loading the orm fully. In this case we should
-            //ignore this error. Moreover if a base is set on the class, it means that the base has been loaded, so
-            //it is NOT a model.
-            // session.send_notification(ShowMessage::METHOD, ShowMessageParams{
-            //     typ: MessageType::ERROR,
-            //     message: "Odoo base models are not found. OdooLS will be unable to generate valid diagnostics".to_string()
-            // });
+        if base_model_syms.is_empty() {
+            // base_model_syms empty so sym cannot be a model, otherwise we would have found it earlier
             return false;
         }
-        if Rc::ptr_eq(symbol, &base_model_syms[0]) ||
-            Rc::ptr_eq(symbol, &model_syms[0]) ||
-            Rc::ptr_eq(symbol, &transient_syms[0])
+        // Check if the symbol is exactly BaseModel, Model or TransientModel
+        // BaseModel, Model, or TransientModel are abstract base classes, we don't want to mark them as models
+        if Rc::ptr_eq(symbol, &base_model_syms[0])
+        || model_syms.first().is_some_and(|s| Rc::ptr_eq(symbol, s))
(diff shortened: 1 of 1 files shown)


server/src/core/python_odoo_builder.rs
```diff
@@ -409,16 +409,16 @@ impl PythonOdooBuilder {
...
+            || model_syms.first().map_or(false, |first| Rc::ptr_eq(symbol, first))
+            || transient_syms.first().map_or(false, |first| Rc::ptr_eq(symbol, first))
         {
             return false;
         }
         if compare_semver(session.sync_odoo.full_version.as_str(), "19.1") >= Ordering::Equal{
             let cached_model_tree = (vec![Sy!("odoo"), Sy!("orm"), Sy!("models_cached")], vec![Sy!("CachedModel")]);
             let cached_model = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &cached_model_tree, u32::MAX);
-            if cached_model.is_empty() || Rc::ptr_eq(symbol, &cached_model[0]){
+            if cached_model.first().map_or(false, |first| Rc::ptr_eq(symbol, first)){
```
cammarosano: Perso, I find this hard to read (although it's clever).
```suggestion
           // probably rename `cached_model` a few lines above to `chached_model_syms`, to be coherent with the other ones (base_model_syms, ...), and to avoid the shadowing below.
            let if Some(chached_model) = chached_model.first() {
              if Rc::ptr_eq(symbol, chached_model)) {
                return false;
              }
            }
```

If you like this suggestion, consider changing the similar pattern above.

server/src/core/python_odoo_builder.rs
```diff
@@ -398,7 +398,7 @@ impl PythonOdooBuilder {
         let base_model_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &base_model_tree, u32::MAX);
         let model_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &model_tree, u32::MAX);
         let transient_syms = session.sync_odoo.get_symbol(session.sync_odoo.config.odoo_path.as_ref().unwrap(), &transient_tree, u32::MAX);
-        if base_model_syms.is_empty() || model_syms.is_empty() || transient_syms.is_empty() {
+        if base_model_syms.is_empty(){
```
cammarosano: I don't understand what you are trying to achieve here...

Anyway, if this is necessary, the first check in the `if` below is redundant (we know `base_model_syms` it's not empty).

cammarosano (changes requested):
If this is a fix, then I suggest you include in the commit message what you are fixing...
(I suppose this function returns false when it should not, in case CachedModel is not there).

Which leads to my next question: do we really need to support branches that haven't rebased yet into master?

server/src/core/python_odoo_builder.rs
mmahrouss: Ah yea, I tried explaining it on discord. 

So why do we return false here? 

1. We have not processed the base model(s)
2. And if we have not processed base model, it means also we are not inheriting from one of them, which if we did, we will trigger building them before the class itself. 

However, if we have a class that inherits Model, and we have not built TransientModel, then it is actually okay, because they are still a model, and maybe we did not build TransientModel because it is still in the queue. 

The only way we are sure we are not a model due to the base model not being loaded, is if BaseModel is not evaluated. 

Historically it was okay because they were all in the same file, so they should all be there at the same time, but not transient_model is in a different file. 

So the change is to only return false if we are missing BaseModel, and basically allow TransientModel and CachedModel to not exist, but still be on a Model ourselves.

And also, we really do not want to ever break early on CachedModel, we just want to now if we are processing `CachedModel` itself or not, because it is an abstract class, not a real odoo model. 

Anyway we could discuss more in office today

server/src/core/python_odoo_builder.rs
mmahrouss: I think the suggestion is good here, but it does not work well if we want to do the same check 2 times as we have above. 

What I suggest I do, slightly more readable is to use `cached_model.first().is_some_and(|first| Rc::ptr_eq(symbol, first))`  hoping that `is_some_and` is better than `map_or(false,...)`

cammarosano (approved):

PR #509, opened 2025-12-11
[FIX] correctly identify non-registry model classes

mmahrouss:
We are trying to do that by checking _register = False.
The check was failing because we handled only one evaluation.
However, in latest versions we also have a type hint, which is its own evaluation.
For that, we implement here a smarter search that only looks are evaluations with values.
In that case, we should be more generic in catching cases.

For later: Consider using this as the check, instead of checking symbol trees


cammarosano (approved):
Very nice!

PR #510, opened 2025-12-11
Fix panicking on models defined outside an odoo module + refactor the is model check

mmahrouss:


mmahrouss:
Check that _register = False, is there for all supported version.
While this is nice and concise on master, we should make sure it is still backwards-compatible. 

Edit: it exists since 14.0, so that check should be correct, if it breaks, then the tests will also fail.

fda-odoo (approved):
Cool ! it should improve performances too ! Looks good to me !

PR #511, opened 2025-12-12
[FIX] [IMP] Resolve property type

cammarosano:
This PR introduces:
- a fix for evaluations that got incorrectly dropped by `follow_ref`
- support for property (functions) in `follow_ref`, following its evaluation to its return type

### DIFF
--- a/server/src/core/symbols/symbol.rs
+++ b/server/src/core/symbols/symbol.rs
@@ -2812,7 +2812,7 @@ impl Symbol {
     ) -> (Vec<Rc<RefCell<Symbol>>>, Vec<Diagnostic>) {
         let mut result: Vec<Rc<RefCell<Symbol>>> = vec![];
         let mut visited_symbols: PtrWeakHashSet<Weak<RefCell<Symbol>>> = PtrWeakHashSet::new();
-        let mut extend_result = |syms: Vec<Rc<RefCell<Symbol>>>| {
+        let mut extend_result = |syms: Vec<Rc<RefCell<Symbol>>>, result: &mut Vec<Rc<RefCell<Symbol>>>, visited_symbols: &mut PtrWeakHashSet<Weak<RefCell<Symbol>>>| {
             syms.iter().for_each(|sym|{
                 if !visited_symbols.contains(sym){
                     visited_symbols.insert(sym.clone());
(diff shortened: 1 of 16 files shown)


odoostan:
Hi @cammarosano long time no see ! I think all "references" features are missing for now ? 
For me "Go to definition" goes to the og file (but the first line not the line where the method is defined)
And "Go to references" and "Find all references" never worked for any method.

server/src/core/symbols/symbol.rs
```diff
@@ -2198,121 +2197,134 @@ impl Symbol {
...
+                        // keep current evaluation
+                        results.push(current_eval);
+                    } else {
+                        // enqueue evaluations to follow, replacing current evaluation
+                        work_queue.extend(next_sym_refs);
+                    }
+                    // TODO: this arm can be easily merged with the one for SymType::CLASS, but
+                    // - do we need to propagate instance = True here too (like we do for variables, but not class...)?
+                    // - do we need to build_now for functions like we do for variables? In case yes, BuildSteps::ARCH_EVAL or VALIDATION?
+                    // - stop_on_type, stop_on_value and max_scope apply also for this case?
```
cammarosano, line 2319: To be discussed with @fda-odoo

cammarosano:
> Hi @cammarosano long time no see ! I think all "references" features are missing for now ? For me "Go to definition" goes to the og file (but the first line not the line where the method is defined) And "Go to references" and "Find all references" never worked for any method.

Hey @odoostan ! Good to hear from you!
Indeed, "references" feature is not yet implemented for Python.
Now, "go to definition" should work.  Can you provide an example where it doesn't?

odoostan:
Hello @cammarosano 
Thank you for following up on this, and for your work on this amazing project in general :+1: 

After some testing, it seems "Go to definition" works fine in all cases except when used in code outside of the usual addons_path.

Example: `base/models/res_company.py`

1. Methods in same module like `button_immediate_install`: work fine :heavy_check_mark: 
2. Generic methods like `browse`: no definition found :x: 

And in the usual paths, we get kind of the opposite.

Example: `account/models/partner.py`

1. Most methods work fine, even generic methods like `browse` :heavy_check_mark: 
2. Methods defined outside of the path like `button_immediate_install`: no definition found :x:

cammarosano:
Thank you @odoostan ! We'll investigate it. Don't hesitate to point out any other issues you might find.

cammarosano:
1st commit was moved to https://github.com/odoo/odoo-ls/pull/520.

cammarosano:
closing in favor of https://github.com/odoo/odoo-ls/pull/700

PR #515, opened 2025-12-18
[FIX] handle pos arg for inverse name on o2m fields

mmahrouss:
Before, we were handling features for inverse name only when it is a keyword argument, but not when it is a positional one this commit aims to fix that

Fixes hover, go to def, and completion.

### DIFF
--- a/server/src/features/completion.rs
+++ b/server/src/features/completion.rs
@@ -588,61 +588,98 @@ fn complete_call(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>, expr_cal
         return complete_expr( &expr_call.func, session, file, offset, is_param, expected_type);
     }
     let scope = Symbol::get_scope_symbol(file.clone(), offset as u32, is_param);
+    let from_module = file.borrow().find_module().clone();
     AstUtils::build_scope(session, &scope);
     let callable_evals = Evaluation::eval_from_ast(session, &expr_call.func, scope, &expr_call.func.range().start(), false, &mut vec![]).0;
     let callable_eval_sym_ptrs = callable_evals.iter().flat_map(|callable_eval|
         Symbol::follow_ref(&callable_eval.symbol.get_symbol(session, &mut None, &mut vec![], None), session, &mut None, false, false, None, None)
     ).collect::<Vec<_>>();
     if let Some((arg_index, arg)) = expr_call.arguments.args.iter().find_position(|arg|
-        offset > arg.range().start().to_usize() && offset <= arg.range().end().to_usize()
-    ){
-        for callable_eval in callable_evals.iter() {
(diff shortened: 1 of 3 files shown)


fda-odoo:
I move it to 1.3.0. I think it's more an [IMP] than a [FIX]

mmahrouss:
> I move it to 1.3.0. I think it's more an [IMP] than a [FIX]

There is a small fix also in `get_indexed_arg_in_call` 
Would you like to maybe split it into two PRs?

fda-odoo:
Yes, we can include the fix in 1.1.3

mmahrouss:
Depends on https://github.com/odoo/odoo-ls/pull/530

server/src/features/completion.rs
```diff
@@ -588,61 +588,98 @@ fn complete_call(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>, expr_cal
...
+            // Relational fields first argument is a model name.
+            if callable_sym.borrow().typ() == SymType::CLASS
+            && arg_index == 0
+            && callable_sym.borrow().is_specific_field_class(session, &["Many2one", "One2many", "Many2many"]) {
+                    return complete_expr(arg, session, file, offset, is_param, &vec![ExpectedType::MODEL_NAME]);
+            }
+            // if class get __init__ method, we need to get the argument from there
+            let func_rc = if callable_sym.borrow().typ() == SymType::CLASS {
+                let class_sym = callable_sym.borrow();
+                if let Some(init_method) = class_sym.get_member_symbol(session, &S!("__init__"), from_module.clone(), false, false, false, false, false).0.first().cloned() {
```
fda-odoo: you coud set only_method to true?

server/src/features/completion.rs
```diff
@@ -588,61 +588,98 @@ fn complete_call(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>, expr_cal
...
+                    init_method
+                } else {
+                    continue;
+                }
+            } else {
+                callable_sym.clone()
+            };
+            let is_on_instance = if callable_sym.borrow().typ() == SymType::CLASS {
+                Some(true)
```
fda-odoo, line 626: false? A class instantiation is not an instance, and the usage of `is_on_instance` to search for arg count will try to pass a `self`argument

server/src/features/features_utils.rs
```diff
@@ -250,39 +251,76 @@ impl FeaturesUtils {
...
+                    continue;
+                }
+            } else {
+                callable_sym.clone()
+            };
+            let func = func_rc.borrow();
+
+            let is_on_instance = if callable_sym.borrow().typ() == SymType::CLASS {
+                // skip self on __init__
+                Some(true)
```
fda-odoo, line 298: false, then?

fda-odoo (changes requested):

fda-odoo (approved):

PR #516, opened 2026-01-09
[FIX] missing SELF evaluation to search method

cammarosano:
The base model's search method should have an evaluation of EvaluationSymbolPtr::SELF.

This used to be the case, but it got lost in a refactor ([1]).

[1]: https://github.com/odoo/odoo-ls/commit/ca26c3e9f60da0ea13da1a3da97fa7123ad0e679


mmahrouss (approved):
Nice catch:) sorry it got lost in the refactoring

fda-odoo (approved):

PR #517, opened 2026-01-14
[FIX] handle empty config response from client

mmahrouss:
We were reading client config every time the config file changes. We got crash reports for unwrapping a None for Config. Already, we restart if the config change, so the check is redundant and causes unnecessary errors Additionally, it is safer to return and Err instead of calling unwrap

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -67,6 +67,7 @@ pub struct SyncOdoo {
     pub full_version: String,
     pub python_version: Vec<u32>,
     pub config: ConfigEntry,
+    pub selected_config_profile: Option<String>,
     pub config_file: Option<ConfigFile>,
     pub config_path: Option<String>,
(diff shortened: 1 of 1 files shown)


mmahrouss:
It is weird though, because we do see the response arriving okay to the client in the error logs, yet we still crash on unwrap, which is unexplainable to me. 

example crash report number 1915 and more

fda-odoo:
That should indeed fix the crash, but isn't it weird that the user select a profile, but in the backend it's the default one that is run because we couldn't get the profile name? (I don't know how we reproduce that)

mmahrouss:
> That should indeed fix the crash, but isn't it weird that the user select a profile, but in the backend it's the default one that is run because we couldn't get the profile name? (I don't know how we reproduce that)

Yes it is indeed true, but it if all the communications go well, this should not happen. 
Also, when the user switches profile, the client already restarts the server, right? so it should be ok

PR #519, opened 2026-01-15
[IMP] Implement Phase 1 of Go To References (#466)

### LINKED ISSUE #466: Field/variable/function references not working
**Your setup**
Version: 1.0.4
Operating System: Ubuntu 24.04
IDE and/or Integration tool (for example: Vscode - official extension): Neovim 0.11

**Describe the bug**
"Go To References" does not appear to be implemented for the lsp. 
**To Reproduce**
Go to any function in odoo and try to use your Editor's go to ref functionality, it will load for a while and then fail.
**Expected behavior**
Go to ref should work

**Additional context**
LSP refactoring like renaming variables doesn't work as well, but I'd assume gotoref is necessary to solve that.

MihaiStreames:
## Summary
- Implements same-file Go To References for Python files
- Adds scope-aware symbol validation to handle variable shadowing
- Includes tests for classes, variables, imports, and self parameter

Closes #466 (partially - Phase 1 only)

## Future work
See TODOs in `server/src/features/references.rs`:
- Cross-file references within module
- Workspace-wide references
- Odoo-specific: XML field refs, string-based model refs


server/src/features/references.rs
```diff
@@ -1,19 +1,117 @@
...
+    /// Same-file references only, local vars, params, methods
+    /// TODO: Cross-file references within module
+    /// TODO: All files within workspace
+    /// TODO: Odoo specific (XML field refs, string-based model refs)
+    pub fn get_references(session: &mut SessionInfo, file_symbol: &Rc<RefCell<Symbol>>, file_info: &Rc<RefCell<FileInfo>>, line: u32, character: u32) -> Option<Vec<Location>> {
+        let offset = file_info.borrow().position_to_offset(line, character);
+
+        let (analyse_ast_result, _range, _call_expr) = AstUtils::get_symbols(session, file_symbol, file_info, offset as u32);
+
+        if analyse_ast_result.evaluations.is_empty() {
```
fda-odoo, line 64: This test seems wrong, but I don't know what you are trying to prevent.
This 'if' will prevent any variable without evaluation, classes or function to get the 'reference' feature working.

server/src/features/references.rs
```diff
@@ -1,19 +1,117 @@
...
+            return None;
+        }
+
+        let eval = &analyse_ast_result.evaluations[0];
```
fda-odoo, line 68: I think there is a misunderstanding on the meaning of "evaluations".
An evaluation is the different values a variable/function can take (inferred with code or signature).
For example:
```python
a = 5
def test():
    return a
```
the variable "a" will have an evaluation to the value "5" (of type int)
and the function "test" will have an evaluation to "a" (that has the previous evaluation to "5")

Taking the evaluation[0] here would mean that you are trying to find all references to "5" if you do it on "a".

server/src/features/references.rs
```diff
@@ -1,19 +1,117 @@
...
+        }
+
+        let file_path = file_symbol.borrow().paths()[0].clone();
+        let mut locations = Vec::new();
+
+        for match_range in name_matches {
+            // infer name
+            let match_offset = match_range.start().to_u32();
+            let scope = Symbol::get_scope_symbol(file_symbol.clone(), match_offset, false);
+            AstUtils::build_scope(session, &scope);
```
fda-odoo, line 92: That's a nice thing to do here!

fda-odoo (commented):
Thank you very much for your pull request, it's a pleasure to see people suggesting new features or functionnalities!

I checked rapidly what you did as this is a "part 1" PR, to be sure you are on the right track! 
And indeed it seems really good. I put some comments about things I'm not sure about, but the execution is efficient!

I guess moving to the "part 2" and searching in other files will be more difficult, if you don't want to browse all files. The idea that comes in my mind would be to only execute your algorithm in files that are importing the current file. Because if they do, they maybe use your symbol.
By chance, each file already knows who is importing them, by the "dependents" variable. You'll see that this field is sorting files in function of the depending step:
ARCH, ARCH_EVAL or VALIDATION.
I think (but I can miss something here) that you should only take files in the ARCH (but it would be easy to change if I forget something)

If you have any question, I'll stay available !

fda-odoo:
@MihaiStreames Hello!
I don't know if you tried to work on the Part 2 already or not, but this feature was planned for the next version 1.3, so if I'll take the development of it from now on. If you have already some work done, do not hesitate to share it, it would be a pleasure to discuss on it.
I'll start from your commit to include it in the final feature !
Thank you again for your input on this!

fda-odoo:
New Pull request: https://github.com/odoo/odoo-ls/pull/531

MihaiStreames:
Hey @fda-odoo! Sorry for the late reply, was on vacation and then swamped with exams.

Thanks for the review and for taking this forward! Your comments make total sense:

1. The `evaluations.is_empty()` check was too aggressive, you're right it would block classes/functions without evaluations
2. Good catch on `evaluations[0]`, I see now that gives you the evaluated *value* rather than the symbol itself. I was treating it like the Definition feature does but didn't fully understand the evaluation semantics

Glad you're building on this for 1.3.0! I don't have any Part 2 work done yet.

Happy to help test or review if useful :+1:

fda-odoo:
Closing for the new PR: https://github.com/odoo/odoo-ls/pull/550

PR #520, opened 2026-01-16
[FIX] follow_ref dropping evaluations incorrectly

cammarosano:
The issue:
In follow_ref's main loop, popping from the front of the results Deque when meaning to replace an evaluation would not necessarily remove the correct one.

In order to fix this and reduce complexity of the main loop with a moving index over a Deque, we switch to a work queue approach:
- evaluations that need to be processed are pushed to and popped from the work queue,
- final evaluations are pushed to the results vector.

Other changes:
- `acc` was renamed to `visited` for clarity.
- some indentation could be reduced with `let else` constructs, improving readability.

### DIFF
--- a/server/tests/test_follow_ref.rs
+++ b/server/tests/test_follow_ref.rs
@@ -1,0 +1,66 @@
+mod setup;
+mod test_utils;
+use odoo_ls_server::core::file_mgr::FileInfo;
+use odoo_ls_server::core::odoo::SyncOdoo;
+use odoo_ls_server::core::symbols::symbol::Symbol;
+use odoo_ls_server::constants::OYarn;
+use odoo_ls_server::threads::SessionInfo;
+use odoo_ls_server::utils::PathSanitizer;
+use odoo_ls_server::Sy;
+use std::cell::RefCell;
+use std::env;
+use std::path::PathBuf;
+use std::rc::Rc;
+use test_utils::get_resolved_symbols_at_position;
+
+#[test]
+fn test_follow_ref() {
+    let (mut odoo, config) = setup::setup::setup_server(false);
+    let mut session = setup::setup::create_init_session(&mut odoo, config);
+    let path = env::current_dir()
+        .unwrap()
+        .join("tests/data/python/expressions/follow_ref.py")
+        .sanitize();
+    setup::setup::prepare_custom_entry_point(&mut session, path.as_str());
+    assert!(session.sync_odoo.entry_point_mgr.borrow().custom_entry_points.len() == 1);
+    let file_mgr = session.sync_odoo.get_file_mgr();
+    let file_info = file_mgr.borrow().get_file_info(&path).unwrap();
+    let file_symbol = SyncOdoo::get_symbol_of_opened_file(&mut session, &PathBuf::from(&path))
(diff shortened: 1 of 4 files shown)


server/tests/test_follow_ref.rs
```diff
@@ -0,0 +1,53 @@
...
+        ("a", (3, 0), "TestClass"),
+        ("b", (4, 0), "TestClass"),
+        ("b", (6, 4), "int"),
+        ("b", (7, 0), "(int | TestClass)"),
+        ("c", (8, 0), "(int | TestClass)"),
+        ("c", (10, 4), "str"),
+        ("c", (11, 0), "(str | int | TestClass)"),
+        ("d", (12, 0), "(str | int | TestClass)"),
+    ] {
+        let hover =
```
fda-odoo: you make your test depend on the hover feature which is useless. If hover results change its display structure, we'll have to update this test.
I think you could compare expected_type with Rc::eq

fda-odoo (changes requested):
Seems good to me ! 
I think it would have been fixable with a pop with index instead of pop_front, but this approach is definitively better :)
Did you test it extensively however? As it's a quite sensible function, let's be sure we are not introducing bugs here

server/tests/test_follow_ref.rs
cammarosano: I have rewritten the test. Can you please check if that's you wanted?

(the force-push diff got mixed with the rebase, sorry).
Actual diff: https://github.com/odoo/odoo-ls/compare/1f75968a52586068087233bc4d1290b14eb62197..861a19cc3bd68cbc6a60ede3f3d0fb4e5d1ba12c#diff-46c5a6f538b509e736ead8b773be9a9604684673cbac6ad9eb0643fa23bdcdea

fda-odoo: This is perfect ! Maybe we could extract this logic (give an offset, and compare evaluation) in a 'util' function, as we need it in other tests.
@mmahrouss is doing the same thing in this pr: https://github.com/odoo/odoo-ls/pull/501/changes#diff-35e6b6fe314bdf5c40586c76d6d93bb5bd5931ec270d2e91f88dfac8d198a688R146

cammarosano:
> Seems good to me ! I think it would have been fixable with a pop with index instead of pop_front, but this approach is definitively better :) 

Thanks!
True, a minimal fix could have been done by popping with the correct index. But one of the goals was to reduce complexity and improve readability, because it's prone to lead to more errors when someone changes this function.
In fact, I spent some time trying to refactor this function, but I stumbled into details I did not fully understand, so I kept it to a minimum. 

>Did you test it extensively however? As it's a quite sensible function, let's be sure we are not introducing bugs here

I tested it, but you might know some cases I could not come up with, so please also give it a go.

server/tests/test_follow_ref.rs
cammarosano: I've factored out the logic for retrieving the resolved symbols into a test util (in a separate commit).
@mmahrouss thought it would be best to leave out the assertions comparing evaluations to expected ones.

fda-odoo (approved):

PR #521, opened 2026-01-16
[FIX] remove wrong stop_on_type in features

mmahrouss:
This is done because functions that have a
return value that refers to a field with a descriptor were not correctly transferred
and were showing Many2One instead of
ModelName

Related to https://github.com/odoo/odoo-ls/pull/520


cammarosano:
image

mmahrouss:
> image

WIll do it today 🙄

PR #524, opened 2026-01-22
[FIX] build get descriptor if needed in next_ref

mmahrouss:
This could happen when building base, that One2many is not built yet, which caused a wrong diagnostic.


fda-odoo (approved):
I was afraid to put on-place rebuild in the follow_ref method, but for the `__get__` and with the guard you added it should be ok, but I guess it won't work everytime, as the file should already be `arch_eval==DONE`

mmahrouss:
> I was afraid to put on-place rebuild in the follow_ref method, but for the `__get__` and with the guard you added it should be ok, but I guess it won't work everytime, as the file should already be `arch_eval==DONE`

Yes indeed it should only run if we are in build_base and fields.py has not ran yet, which is not very common

PR #527, opened 2026-01-23
[IMP] Handle asset odoo nodes

mmahrouss:


server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -603,4 +605,125 @@ impl XmlArchBuilder {
...
     }
+
+    fn load_asset(&mut self, session: &mut SessionInfo, node: &Node, diagnostics: &mut Vec<Diagnostic>) -> bool {
+        if node.tag_name().name() != "asset" { return false; }
+        // Validate required attributes: id, name
+        let mut found_id = None;
+        let mut has_name = false;
+        for attr in node.attributes() {
+            match attr.name() {
+                "id" => { found_id = Some(attr.value().to_string()); },
```
fda-odoo, line 616: maybe we could check that there is not the same attribute multiple times

fda-odoo (approved):

PR #528, opened 2026-01-23
[IMP] sort modules by load order

cammarosano:
Module packages are now built in the same order as the Odoo server loads them.

### DIFF
--- a/server/src/core/module_load_order.rs
+++ b/server/src/core/module_load_order.rs
@@ -1,0 +1,352 @@
+use std::collections::HashMap;
+
+pub struct SortResult<'a> {
+    pub sorted: Vec<&'a str>,
+    pub invalid: Vec<&'a str>, // due to missing direct/indirect dependencies or dependency cycle
+    pub missing: Vec<&'a str>,
+    pub cycles: Vec<Vec<&'a str>>,
+}
+
+/// Sort modules based on Odoo module load order logic.
+/// 
+/// Input: Vec of (module_name, dependencies)
+/// 
+/// Output: SortResult containing sorted valid modules and invalid modules
+/// 
+/// Note: 'base' module is expected to be included in the input modules.
+pub fn sort_by_load_order<'a>(modules: Vec<(&'a str, Vec<&'a str>)>) -> SortResult<'a> {
+    let mut graph = Graph::from(modules);
+    let (sorted, invalid) = graph.get_load_order();
+    let mut missing = Vec::new();
+    let mut cycles = Vec::new();
+    for error in graph.errors {
+        match error {
+            ValidationError::MissingModule(m) => missing.push(m),
+            ValidationError::DependencyCycle(cycle) => cycles.push(cycle),
+        }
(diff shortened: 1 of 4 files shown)


server/src/core/odoo.rs
```diff
@@ -503,6 +508,43 @@ impl SyncOdoo {
...
+            .chain([("base", vec![])]) // Include "base" for proper dependency resolution
+            .collect();
+
+        let sort_result = sort_by_load_order(nodes);
+        debug_assert!(sort_result.sorted.first() == Some(&"base"), "The first module after sorting should be 'base'");
+
+        sort_result.sorted
+            .iter()
+            .skip(1) // skip "base" 
+            // TODO: decide what to do with invalid modules. For now, we append them at the end.
```
cammarosano, line 543: :eyes:

server/src/core/module_load_order.rs
```diff
@@ -0,0 +1,350 @@
...
+        // - recursion_stack is only for cycle detection during traversal
+        // - if a cycle is detected, the module is cached as `false` from the inner recursive call
+        // - once cached, the result is valid for all future lookups regardless of traversal path
+        // Note: the block for cached! macro cannot contain return statements (therefore the indirection)
+        cached!(self.validation_cache, name, {
+            self._is_valid_module(name, recursion_stack)
+        })
+    }
+
+    // a module is valid when:
```
fda-odoo: If you put comments on a function, try to turn it into a documentation string.
Use /* */ instead if // to get docstring autogenerated

```rust
/* Check if a module is valid.
This implementation is different from the Odoo python framework (see
ModuleGraph._update_depends and _update_depth).
A module is valid when:
- it exists
- is not part of a dependency cycle
- all its dependencies are valid modules */
```

server/src/core/module_load_order.rs
```diff
@@ -0,0 +1,350 @@
...
+        };
+        recursion_stack.push(name);
+        let is_valid = dependencies
+            .iter()
+            .all(|&dep| self.is_valid_module(dep, recursion_stack));
+        recursion_stack.pop();
+        is_valid
+    }
+
+    // ===== Sort algo (topological sort) =======
```
fda-odoo, line 151: same here

fda-odoo (changes requested):
Great work! I like the usage of `cached` :)

server/src/core/module_load_order.rs
cammarosano: As this file already had some functions documented with `///`, I went for `///` instead of `/* */`.

server/src/core/module_load_order.rs
cammarosano: this `// ===== Sort algo (topological sort) =======` is just a section divider for better organization of the methods in the Graph struct (so, no doctring). But I added `///` to what follows next, to make it a docstring for the `get_sort_key` method.

fda-odoo (approved):

PR #531, opened 2026-01-28
[IMP] Implement Go To References Part 1 - python code

fda-odoo:
First part of the Go To References.
Include Python searches: python variables, models.
Do not search or work in:
- xml/csv files
- manifest files
- lambda functions


server/src/core/evaluation.rs
```diff
@@ -1266,11 +1267,13 @@ impl Evaluation {
...
-                    },
-                    _ => {}
+                if odoo.evaluation_search.is_some() {
+                    match operator.op {
+                        Operator::Add | Operator::Mult | Operator::Sub | Operator::Div | Operator::BitAnd | Operator::BitOr | Operator::BitXor |Operator::FloorDiv | Operator::LShift | Operator::MatMult | Operator::Mod | Operator::Pow | Operator::RShift => {
+                            Evaluation::eval_from_ast(session, &operator.left, parent.clone(), max_infer, false, required_dependencies);
+                            Evaluation::eval_from_ast(session, &operator.right, parent.clone(), max_infer, false, required_dependencies);
+                        }
+                    }
                 }
```
mmahrouss, line 1277: The match is not needed here, we can drop it
```suggestion
                if odoo.evaluation_search.is_some() {
                    Evaluation::eval_from_ast(session, &operator.left, parent.clone(), max_infer, false, required_dependencies);
                    Evaluation::eval_from_ast(session, &operator.right, parent.clone(), max_infer, false, required_dependencies);
}
```

server/src/core/odoo.rs
```diff
@@ -1468,14 +1604,28 @@ impl Odoo {
                 match ast_type {
```
mmahrouss, line 1604: Suggestion to have less code duplication and readability
```suggestion
                if file_info.borrow().file_info_ast.borrow().indexed_module.is_none() && matches!(ast_type, AstType::Python) {
                    return Ok(None);
                }
                let location_fn = match (ast_type, is_declaration) {
                    (AstType::Python, true) => DeclarationFeature::get_location,
                    (AstType::Python, false) => DefinitionFeature::get_location,
                    (AstType::Xml, true) => DeclarationFeature::get_location_xml,
                    (AstType::Xml, false) => DefinitionFeature::get_location_xml,
                    (AstType::Csv, true) => DeclarationFeature::get_location_csv,
                    (AstType::Csv, false) => DefinitionFeature::get_location_csv,
                };
                return Ok(location_fn(session, &file_symbol, &file_info, params.text_document_position_params.position.line, params.text_document_position_params.position.character));
```

server/src/features/definition.rs
mmahrouss, line 31: could be useful to isolate `get_location_xml` from declaration and definition to goto_utils, to have less code duplication. what do you think?

server/tests/test_references.rs
```diff
@@ -0,0 +1,95 @@
...
+    let mut references = get_references(&mut session, &test_file, Position::new(50, 4));
+    assert_in_result(&mut references, "module_1/models/models.py", 1, 30);
+    assert!(references.len() == 0, "Some references were not expected: {}",
+        references.iter().map(|r| format!("{}:{}:{}", r.uri.as_str(), r.range.start.line, r.range.start.character)).collect::<Vec<String>>().join(", ")
+    );
+    
+    // for r in references.iter() {
+    //     error!("Reference found at {}:{}:{}", r.uri.as_str(), r.range.start.line, r.range.start.character);
+    // }
+}
```
mmahrouss, line 59: unneeded old comment?

server/src/core/symbols/symbol.rs
```diff
@@ -2352,6 +2352,34 @@ impl Symbol {
...
+        symbols.push_back(evaluation.clone());
+        while let Some(current_sym) = symbols.pop_front() {
+            let EvaluationSymbolPtr::WEAK(w) = &current_sym else {
+                res.push(current_sym.clone());
+                continue;
+            };
+            let Some(symbol) = w.weak.upgrade() else {
+                res.push(current_sym.clone());
+                continue;
+            };
```
mmahrouss, line 2367: Why are we adding the sym to res in the else? it is an expired weak, so it is better to just continue, right?

server/src/core/evaluation.rs
```diff
@@ -1351,6 +1357,65 @@ impl Evaluation {
...
             _ => {}
         }
+        if let Some(evaluation_search) = session.sync_odoo.evaluation_search.as_ref() {
+            for eval in evals.iter() {
+                if eval.symbol.sym.is_weak() && let Some(weak) = eval.symbol.sym.as_weak().weak.upgrade() {
+                    if Rc::ptr_eq(&weak, evaluation_search) {
+                        if found_one_reference {
+                            //if we have multiple matches, it means that that ast can reference it multiple times, but we only want to know if that ast matches or not
+                            continue;
+                        }
```
mmahrouss, line 1367: We could move the check of `found_one_reference` and break early to avoid extra checks
```suggestion
            for eval in evals.iter() {
                if found_one_reference {
                    //if we have multiple matches, it means that that ast can reference it multiple times, but we only want to know if that ast matches or not
                    break;
                }
                if eval.symbol.sym.is_weak() && let Some(weak) = eval.symbol.sym.as_weak().weak.upgrade() {
                    if Rc::ptr_eq(&weak, evaluation_search) {
```

server/src/features/references.rs
```diff
@@ -1,19 +1,97 @@
...
 
 impl ReferenceFeature {
-    pub fn get_references(_session: &mut SessionInfo, _file_symbol: &Rc<RefCell<Symbol>>, _file_info: &Rc<RefCell<FileInfo>>, _line: u32, _character: u32) -> Option<Vec<Location>> {
-        // Implementation for getting references
-        None
+    /// Basic find all references implementation
+    /// Same-file references only, local vars, params, methods
+    /// TODO: Cross-file references within module
+    /// TODO: All files within workspace
+    /// TODO: Odoo specific (XML field refs, string-based model refs)
```
mmahrouss, line 25: This comments are no longer correct, no? perhaps only XML refs are missing?

server/src/features/references.rs
```diff
@@ -67,4 +145,206 @@ impl ReferenceFeature {
...
+    pub fn visit_vec_stmt(&mut self, session: &mut SessionInfo, vec_ast: &Vec<Stmt>) {
+        for stmt in vec_ast.iter() {
+            match stmt {
+                Stmt::FunctionDef(f) => {
+                    let sym = self.sym_stack.last().unwrap().borrow().get_positioned_symbol(&OYarn::from(f.name.to_string()), &f.range);
+                    if let Some(sym) = sym {
+                        self.sym_stack.push(sym);
+                        self.visit_vec_stmt(session, &f.body);
+                        self.sym_stack.pop();
+                    }
```
mmahrouss, line 170: We could isolate this into its own method `visit_function_def` to be consistent with the other branches, what do you think?

server/src/features/references.rs
```diff
@@ -67,4 +145,206 @@ impl ReferenceFeature {
...
+        };
+        for alias in name_aliases.iter() {
+            if alias.name.id == "*" {
+                continue;
+            }
+            let var_name = if alias.asname.is_none() {
+                S!(alias.name.split(".").next().unwrap())
+            } else {
+                alias.asname.as_ref().unwrap().clone().to_string()
+            };
```
mmahrouss, line 242: Could be a bit more readable to use match here.
```suggestion
            let var_name = match alias.asname.as_ref() {
                None => S!(alias.name.split(".").next().unwrap()),
                Some(asname) => S!(asname),
            };
```

mmahrouss (approved):
Looks good, I added some cosmetic suggestions, nothing functional:D

server/src/core/symbols/symbol.rs
fda-odoo: If we can't go further in the evaluations, we want to still keep the actual result, so the else is adding it to the list. Seems correct to me.

server/src/core/symbols/class_symbol.rs
```diff
@@ -84,7 +85,7 @@ impl ClassSymbol {
     }
 
     pub fn add_symbol(&mut self, content: &Rc<RefCell<Symbol>>, section: u32) {
-        let sections = self.symbols.entry(content.borrow().name().clone()).or_insert(HashMap::new());
+        let sections = self.symbols.entry(content.borrow().name().clone()).or_insert(HashMap::default());
```
cammarosano, line 88: or simply `or_default()`

```suggestion
        let sections = self.symbols.entry(content.borrow().name().clone()).or_default());
```

(same for function_symbol, file_symbol, etc)

server/src/utils.rs
```diff
@@ -374,3 +374,27 @@ macro_rules! warn_or_panic {
...
     }
 }
+
+use std::hash::{BuildHasherDefault, Hasher};
+#[derive(Default)]
+pub struct U32IdentityHasher(u64);
+
+impl Hasher for U32IdentityHasher {
+	fn write(&mut self, bytes: &[u8]) {
+		// fallback lent (rare si on n'utilise que u32)
```
cammarosano, line 384: 🥖

server/src/utils.rs
```diff
@@ -374,3 +374,27 @@ macro_rules! warn_or_panic {
         }
     }
 }
+
+use std::hash::{BuildHasherDefault, Hasher};
+#[derive(Default)]
+pub struct U32IdentityHasher(u64);
+
+impl Hasher for U32IdentityHasher {
```
cammarosano, line 382: This looks cool!
But did you perf? Did it improve?

Have you considered using a Vector instead of Hashmap? If the idea is to use small sequential integers as keys, wouldn't a Vector be simpler (and cache friendlier)?
You could just type alias it to "SymbolMap" or something, ex:
```rust
type SymbolMap = Vec<u32, Vec<Rc<RefCell<Symbol>>>>
```
or maybe wrap the inner Vec in an Option (to signal the absence/presence of entry for a given key), replicating what a map would return for `get`:
```rust
type SymbolMap = Vec<u32, Option<Vec<Rc<RefCell<Symbol>>>>>
```

Your optimization makes the hashmap behave like a vector (no hashing, direct indexing), so why not go for a vector?

server/src/core/evaluation.rs
```diff
@@ -282,7 +283,7 @@ impl Evaluation {
         Evaluation {
             symbol: EvaluationSymbol {
                 sym: EvaluationSymbolPtr::WEAK(EvaluationSymbolWeak{
-                    weak: Rc::downgrade(&odoo.get_symbol("", &(vec![Sy!("builtins")], vec![Sy!("list")]), u32::MAX).last().expect("builtins list not found")),
+                    weak: odoo.get_ts_list(),
```
cammarosano, line 286: Very minor:
You replaced this `expect` with a `unwrap` (inside `get_ts_list`). I would suggest keeping the `expect` with an informative message.
(same for the other `get_ts_*`s below)
Also, see other comment about using a macro. That could be a way to bring the `expect` back.

server/src/core/odoo.rs
```diff
@@ -1204,6 +1248,90 @@ impl SyncOdoo {
...
+        self.typeshed_weak_cache.bytes.clone()
+    }
+
+    pub fn get_ts_object(&mut self) -> Weak<RefCell<Symbol>> {
+        if self.typeshed_weak_cache.object.is_expired() {
+            self.typeshed_weak_cache.object = Rc::downgrade(&self.get_symbol("", &(vec![Sy!("builtins")], vec![Sy!("object")]), u32::MAX).last().unwrap().clone());
+        }
+        self.typeshed_weak_cache.object.clone()
+    }
+
```
cammarosano, line 1334: Maybe a macro?
```rust
  macro_rules! typeshed_getter {                                                                                                                           
      ($fn_name:ident, $field:ident, $($path:expr),+) => {                                                                                               
          pub fn $fn_name(&mut self) -> Weak<RefCell<Symbol>> {                                                                                            
              if self.typeshed_weak_cache.$field.is_expired() {                                                                                            
                  self.typeshed_weak_cache.$field = Rc::downgrade(
                      &self.get_symbol("", &(vec![Sy!("builtins")], vec![Sy!($($path),+)]), u32::MAX)
                          .last()
                          .expect(concat!("builtins ", stringify!($field), " not found"))
                  );
              }
              self.typeshed_weak_cache.$field.clone()
          }
      };
  }

  impl SyncOdoo {
      typeshed_getter!(get_ts_dict,     dict,     "dict");
      typeshed_getter!(get_ts_tuple,    tuple,    "tuple");
      typeshed_getter!(get_ts_set,      set,      "set");
      typeshed_getter!(get_ts_list,     list,     "list");
      typeshed_getter!(get_ts_string,   string,   "str");
      typeshed_getter!(get_ts_boolean,  boolean,  "bool");
      typeshed_getter!(get_ts_int,      int,      "int");
      typeshed_getter!(get_ts_float,    float,    "float");
      typeshed_getter!(get_ts_complex,  complex,  "complex");
      typeshed_getter!(get_ts_ellipsis, ellipsis, "Ellipsis");
      typeshed_getter!(get_ts_bytes,    bytes,    "bytes");
      typeshed_getter!(get_ts_object,   object,   "object");
  }
```

server/src/features/references.rs
```diff
@@ -81,6 +88,9 @@ impl ReferenceFeature {
 
     fn references_in_file(session: &mut SessionInfo, file_symbol: &Rc<RefCell<Symbol>>, file_info: &Rc<RefCell<FileInfo>>, symbol_name: &String, target_symbol_rc: &Rc<RefCell<Symbol>>) -> Vec<Location> {
         error!("Searching references for symbol {} in file {}", symbol_name, file_symbol.borrow().paths()[0]);
+        if file_symbol.borrow().paths()[0].contains("test_event_sale_with_product_configurator") {
+            error!("found one");
+        }
```
cammarosano, line 93: I'm not sure if you're into cleaning up the individual commits... but this looks weird. 😄 
(I know this gets removed in a later commit)

server/src/core/odoo.rs
```diff
@@ -1426,10 +1554,18 @@ impl Odoo {
...
 
     pub fn handle_goto_definition(session: &mut SessionInfo, params: GotoDefinitionParams) -> Result<Option<GotoDefinitionResponse>, ResponseError> {
+        return Odoo::handle_gotos(session, params, false)
+    }
+
+    pub fn handle_goto_declaration(session: &mut SessionInfo, params: GotoDefinitionParams) -> Result<Option<GotoDeclarationResponse>, ResponseError> {
+        return Odoo::handle_gotos(session, params, true)
+    }
+
+    fn handle_gotos(session: &mut SessionInfo, params: GotoDefinitionParams, is_declaration: bool) -> Result<Option<GotoDeclarationResponse>, ResponseError> {
```
cammarosano, line 1564: Perso, I would use an enum with 2 goto types (definition and declaration) instead of this `is_declaration` boolean.

server/src/core/symbols/symbol.rs
```diff
@@ -2352,6 +2352,34 @@ impl Symbol {
         results
     }
 
+    pub fn follow_imported_ref(evaluation: &EvaluationSymbolPtr, session: &mut SessionInfo, context: &mut Option<Context>) -> Vec<EvaluationSymbolPtr> {
+        let mut res = vec![];
+        let mut symbols = VecDeque::new();
+        symbols.push_back(evaluation.clone());
+        while let Some(current_sym) = symbols.pop_front() {
```
cammarosano, line 2359: 🤩

server/src/core/symbols/symbol.rs
cammarosano: I was about to comment the same thing. Keeping an EvaluationSymbolPtr::WEAK to an expired weak is useful? Why?

server/src/core/evaluation.rs
```diff
@@ -1351,6 +1357,65 @@ impl Evaluation {
...
+                                                        uri: uri,
+                                                        range: transformed_range,
+                                                    });
+                                                }
+                                            }
+                                        }
+                                    }
+                                },
+                                _ => {}
+                            }
```
cammarosano, line 1412:
```
Every time
    you write code like this
        a puppy
            dies
                 !!!!
```
Image

😆 

I suggest replacing this pattern
```
match value {
   the_one_pattern(x) => {...}
   _ => {}
}
```
By this:
```
let the_one_pattern(x) = value else {
     continue;
}
...
```
The same for `if let` without any code after it -> `let ... else { continue }`

server/src/core/evaluation.rs
```diff
@@ -1351,6 +1357,65 @@ impl Evaluation {
             }
             _ => {}
         }
+        if let Some(evaluation_search) = session.sync_odoo.evaluation_search.as_ref() {
```
cammarosano, line 1360: Beware of early returns in the code above this part!
A possible way to ensure this part always runs would be to make this a sibling function of current `analyze_ast`:
```rust
fn analyze_ast() -> result {
   result = old_analyze_ast;
   new_function_for_references(result.evals);
   return result;
}
```
And then `found_one_reference` would be declared inside the`new_function_for_references`

Edit: maybe make `new_function_for_references` get called by `eval_from_ast`, right after the call to `analyze_ast`?

server/tests/test_references.rs
```diff
@@ -0,0 +1,95 @@
...
+        references.iter().map(|r| format!("{}:{}:{}", r.uri.as_str(), r.range.start.line, r.range.start.character)).collect::<Vec<String>>().join(", ")
+    );
+
+    //reference of a simple variable
+    let mut references = get_references(&mut session, &test_file, Position::new(50, 4));
+    assert_in_result(&mut references, "module_1/models/models.py", 1, 30);
+    assert!(references.len() == 0, "Some references were not expected: {}",
+        references.iter().map(|r| format!("{}:{}:{}", r.uri.as_str(), r.range.start.line, r.range.start.character)).collect::<Vec<String>>().join(", ")
+    );
+
```
cammarosano, line 55: I suggest adding more tests 😬 
Like functions/methods, shadowing/nested scopes... or when the lack of results is expected (like a literal).

server/src/features/references.rs
```diff
@@ -67,4 +145,206 @@ impl ReferenceFeature {
     pub fn get_references_csv(_session: &mut SessionInfo, _file_symbol: &Rc<RefCell<Symbol>>, _file_info: &Rc<RefCell<FileInfo>>, _line: u32, _character: u32) -> Option<Vec<Location>> {
         None
     }
+}
+
+struct ReferenceVisitor {
+    sym_stack: Vec<Rc<RefCell<Symbol>>>,
+}
+
+impl ReferenceVisitor {
```
cammarosano, line 154: this is beautiful 💟

server/src/core/evaluation.rs
```diff
@@ -1351,6 +1357,65 @@ impl Evaluation {
...
+                        let file_info = session.sync_odoo.get_file_mgr().borrow().get_file_info(&file.borrow().paths()[0]);
+                        if let Some(file_info) = file_info {
+                            found_one_reference = true;
+                            let range= ast.range();
+                            let transformed_range = file_info.borrow().text_range_to_range(&range, session.sync_odoo.encoding);
+                            //check that a previous call to analyze_ast (recursively for ex) didn't already add this expression
+                            let uri = FileMgr::pathname2uri(&file.borrow().paths().first().unwrap());
+                            if session.sync_odoo.evaluation_locations.is_empty() ||
+                            session.sync_odoo.evaluation_locations.last().unwrap().uri != uri ||
+                            session.sync_odoo.evaluation_locations.last().unwrap().range.start.line != transformed_range.start.line{
```
cammarosano, line 1378: Maybe add a comment explaining what this deduplication of consecutive matches solves? (`env["some_model"].sudo().browse()` etc)

cammarosano (approved):
Very nice!! Killer feature, nice approach!
I have added mostly minor style suggestions, feel free to ignore them.

andg-odoo:
Hey! :wave:
I did a quick test checking out this branch and just working as normal. It's great! 99% of the time it works perfectly with what I would use it for.

However, there is small areas that I'm not sure are broken due to the unimplemented list above in the description or something else.

Here is an example of a function that partially works:
```py
def _log(self, uuid, account_id, data):
  purpose = data['endpoint']
  account_name = account_id.display_name if account_id else 'N/A'
  self._create_account(uuid, account_id, data)
  
  _logger.info('Request on behalf of %s for api_id %s to %s', uuid, account_name, purpose)
  
  http.request.env['proxy.log'].sudo().create({
    'uuid': uuid,
    'account_id': account_id.id if account_id else False,
    'purpose': purpose,
    'is_paid': bool(self._check_paid(data)),
  })
```

Doing a get reference call on the variables will get the following info:
- `uuid` will find only the call in the parameter to `self._create_account`.
- `account_id` will only find the calls on the assigning for `account_name` and `self._create_account`, but not the `account_id` inside of the `create()` call.
- `data` will only find the calsl to assign `purpose` and `self._create_account`, but not the call inside of `check_paid`.

It seems like it can find use in top level calls but nothing within parameters of function calls that aren't on self.

Goto def appears to work for all variables seen here even in areas get refs fail.

Obviously this isn't merged yet so could still change but just wanted to put my experience in for the implementation.

Thanks again, watching the growth of this has been great and excited for this feature once it is merged :smile:

fda-odoo:
Hey @andg-odoo ,

Thank you for your feedback :) Indeed this PR is a 'part 1' to review and validate the architecture of the feature, but it is still missing some important things. And it's pointing us to some bugs too :D 
Aside of xml_id, manifests and lambda that are missing here, the branch is pointing out that some part of arguments are not processed by the tool. that's the case for some other expressions too. That's will be indeed part of the task once it's stabilized. I hope we'll be able to release it soon :)

fda-odoo:
Closed for https://github.com/odoo/odoo-ls/pull/550

PR #538, opened 2026-02-03
[FIX] report error not crash on invalid file states

mmahrouss:
This can happen when we get didChange without a didOpen coming before it, which is not allowed by the LSP. However, we saw multiple crash reports having that state.

This PR aims at showing a warning message to the user, instead of panicking and crashing the server

### DIFF
--- a/server/src/core/file_mgr.rs
+++ b/server/src/core/file_mgr.rs
@@ -144,14 +144,19 @@ impl FileInfo {
             _ => {},
         }
         self.diagnostics.clear();
-        if let Some(content) = content {
+        if let (Some(content), Some(version)) = (content, version) {
             // If we are in did open, we create a new text_document
             // I.E. we have one content change event with no range
             // See [`Odoo:handle_did_open`]
             if content.len() == 1 && content[0].range.is_none() {
                 self.file_info_ast.borrow_mut().text_document = Some(TextDocument::new(content[0].text.clone(), self.version.expect("Expected version on file did Open")));
             } else {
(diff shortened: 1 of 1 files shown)


server/src/core/file_mgr.rs
```diff
@@ -144,14 +144,19 @@ impl FileInfo {
...
             // I.E. we have one content change event with no range
             // See [`Odoo:handle_did_open`]
             if content.len() == 1 && content[0].range.is_none() {
                 self.file_info_ast.borrow_mut().text_document = Some(TextDocument::new(content[0].text.clone(), self.version.expect("Expected version on file did Open")));
             } else {
-                self.file_info_ast.borrow_mut().text_document.as_mut().unwrap().apply_changes(content.clone(), version.unwrap(), session.sync_odoo.encoding);
+                if let Some(text_document_mut) = self.file_info_ast.borrow_mut().text_document.as_mut(){
+                    text_document_mut.apply_changes(content.clone(), version, session.sync_odoo.encoding);
+                } else {
+                    session.show_message(MessageType::WARNING, format!("Invalid file state at: {}, consider reloading the file or restarting the server", path));
```
fda-odoo, line 157: are we sure that this state happen because we got a DidChange before a DidOpen, or is it because we removed the FileInfo of an opened file?

mmahrouss: But we only set text_document to false when the file is not opened and file_cache is false, no? 

Or if the file `is_external` and not in workspace, we do it after validation, and if it is external we will not get did change, did open, or similar requests for them anyway

fda-odoo:
Should never happen. If this code is reached, we should not hide the panic

PR #539, opened 2026-02-04
[FIX] No panic on an old open request for __init__

mmahrouss:
Before this was causing panic, while it is safe to just ignore it

This PR, makes it panic only on unreachable cases, otherwise, silently leave. It is safe to not process a deleted `__init__,py`

### DIFF
--- a/server/src/core/entry_point.rs
+++ b/server/src/core/entry_point.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, cmp, collections::HashMap, path::PathBuf, rc::{Rc, Weak}, u32};
+use std::{cell::RefCell, cmp, collections::HashMap, path::{Path, PathBuf}, rc::{Rc, Weak}, u32};
 
 use tracing::{error, info, warn};
 use weak_table::PtrWeakHashSet;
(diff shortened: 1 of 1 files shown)


server/src/core/entry_point.rs
```diff
@@ -201,9 +201,13 @@ impl EntryPointMgr {
                         warn!("new custom entry point for manifest without related init.py is not supported outside of main entry point. skipping...");
                         session.sync_odoo.entry_point_mgr.borrow_mut().remove_entries_with_path(tree_path);
                     } else {
-                        // There was an __init__.py, that was renamed or deleted.
-                        // Another notification will come for the deletion of the file, so we just warn here.
-                        warn_or_panic!("Trying to create a custom entrypoint on a namespace symbol: {:?}", new_sym.borrow().paths());
+                        if Path::new(file_path).exists() || !file_path.ends_with("__init__.py") {
+                            // a file exists, or does not end with __init__.py
```
fda-odoo, line 205: are you sure that `!file_path.ends_with("__init__.py")` is the right test?
It's confusing me. If we are on a namespace symbol, I thought you would want to test if an `__init__.py` file exists or not in the namespace. Something like:
`Path::new(file_path.join("__init__.py").exists()`
I don't see the point of `Path::new(file_path).exists()` either

fda-odoo (changes requested):

server/src/core/entry_point.rs
mmahrouss: So the goal of this condition is not panic when we have an `__init__.py` file that does not exist anymore. 

This is the case that we can predict, and is safe to skip, other cases we should still warn or panic, which in theory should be unreachable. 

I was able to replicate the panic, by having a folder ( custom entry points) having `__init__` and `__manifest__` and then trigering a rebuild by editing a file in the orm, while it is reloading, I delete `__init__` and then doing a hover in `__manifest__` 

Then when `create_from_path` gets called during the rebuild, (due to the py package being in self import), then since we check if `__init__.py` exists or not, we get false, so we create a  `namespace` which is wrong. 

But it should be okay, because there should come a didClose and/or didDelete and then we will delete the symbol. 

So that is why we want to only catch the case that there is a file that exists and is not `__Init__` then something else is wrong and we want to catch it. 

Another thing to note is, probably  #541 already will prevent the panic from happenign at all, because usually the symbol will be removed correctly, while it was not before, but since I am not sure about all the cases, it might still be safe to keep the change of the check here?

fda-odoo:
This code should be in unreachable code. If we reach it, we should not hide the panic, as it's an error

PR #542, opened 2026-02-06
[FIX] removing skipping already validation rebuilt

mmahrouss:
This was only rarely occurring, and it was causing an issue because we were re-adding  files in add_dependents_to_validation which is valid for when something in base depends on a model. which is not common enough to cause errors, but leaving it like so can leave some base files not built and stuck and pending


server/src/core/odoo.rs
```diff
@@ -719,6 +718,9 @@ impl SyncOdoo {
             }
             let sym = session.sync_odoo.pop_item(BuildSteps::ARCH);
             if let Some(sym_rc) = sym {
+                if DEBUG_THREADS {
```
fda-odoo: DEBUG_THREAD is meant to debug communication and message order between threads.
To debug steps, you should rather use DEBUG_STEPS

fda-odoo (changes requested):

server/src/core/odoo.rs
mmahrouss: Alright, will update it now

fda-odoo (approved):

PR #548, opened 2026-02-16
[REF] XmlValidator: cache all_fields call

cammarosano:
Performance analysis shows that Symbol::all_fields() is the most expensive function call during record validation. In order to avoid redundant recomputation when multiple records reference the same model in a single XML file, we cache results per model name.

Other changes for performance or code improvements:

- Deduplicate get_main_symbols call in validate_record
- Replace unnecessary diagnostic.clone() with moves
- Comment out no-op add_model_dependencies call (XmlFileSymbol not handled), and removed model_dependencies from method parameters
- Comment out mandatory_fields computation, which is not read anywhere
- Store module on XmlValidator struct, remove from all method parameters
- Rename `validate_xml_id` to `validate_data` for clarity (as we are not validating the XML ID itself, but the whole data entry)

--------------

Measured improvements using tracing-timing (as per alpha-imp_perf_xml_validation-with-instrumentation-rcdl), running odoo_ls_server in --parse mode with community + enterprise 19.2:

**Before**
Timing histogram span="xml_validation" event="close" count=6687 p50_ms=0.147455 p90_ms=0.962559 p99_ms=8.126463 max_ms=149.946367 **total_ms=4089.7898720000026**

**After**
Timing histogram span="xml_validation" event="close" count=6687 p50_ms=0.112127 p90_ms=0.288767 p99_ms=0.724991 max_ms=5.963775 **total_ms=968.1314160000005**

                                                                                             
  | Metric | Before | After | Change |                                                       
  |--------|--------|-------|--------|                                                       
  | count | 6687 | 6687 | — |                                                                
  | p50 (ms) | 0.147 | 0.112 | -24% |                                                        
  | p90 (ms) | 0.963 | 0.289 | -70% |                                                        
  | p99 (ms) | 8.126 | 0.725 | -91% |                                                        
  | max (ms) | 149.95 | 5.96 | -96% |                                                        
  | total (ms) | 4089.8 | 968.1 | **-76%** |

### DIFF
--- a/server/src/core/xml_validation.rs
+++ b/server/src/core/xml_validation.rs
@@ -3,22 +3,27 @@ use std::{cell::RefCell, cmp::Ordering, collections::{HashMap, HashSet}, rc::Rc}
 use lsp_types::{Diagnostic, Position, Range};
 use tracing::{info, trace};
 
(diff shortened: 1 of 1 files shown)


fda-odoo (approved):

cammarosano:
@fda-odoo rebased.

PR #549, opened 2026-02-23
[FIX] follow ref in completion of calls

mmahrouss:
This makes sure we are completing the final callable a class for fields to get the correct result


server/src/features/completion.rs
```diff
@@ -590,67 +590,70 @@ fn complete_call(session: &mut SessionInfo, file: &Rc<RefCell<Symbol>>, expr_cal
...
-                                                    expected_type.push(ExpectedType::CLASS(evaluation.clone()));
-                                                }
-                                            }
-                                        },
-                                        EvaluationSymbolPtr::DOMAIN => {
-                                            if let Some(parent) = callable.context.get(&S!("base_attr"))
-                                                .and_then(|parent_value| parent_value.as_symbol().upgrade()) {
-                                                expected_type.push(ExpectedType::DOMAIN(parent));
+    let callable_eval_sym_ptrs = callable_evals.iter().flat_map(|callable_eval|
+        Symbol::follow_ref(&callable_eval.symbol.get_symbol(session, &mut None, &mut vec![], None), session, &mut None, false, false, None, None)
```
mmahrouss, line 594: This is the main change, the rest is mostly a refacto

fda-odoo (approved):

PR #550, opened 2026-02-24
GoToReference - Python, XML and CSV files

fda-odoo:
Depends on https://github.com/odoo/odoo-ls/pull/531

Implement gotoref for modules and odoo data


server/src/core/odoo.rs
```diff
@@ -1439,10 +1572,18 @@ impl Odoo {
...
+    pub fn handle_goto_declaration(session: &mut SessionInfo, params: GotoDefinitionParams) -> Result<Option<GotoDeclarationResponse>, ResponseError> {
+        return Odoo::handle_gotos(session, params, true)
+    }
+
+    fn handle_gotos(session: &mut SessionInfo, params: GotoDefinitionParams, is_declaration: bool) -> Result<Option<GotoDeclarationResponse>, ResponseError> {
         if session.sync_odoo.state_init == InitState::NOT_READY {
             return Ok(None);
         }
-        session.log_message(MessageType::INFO, format!("GoToDefinition requested on {} at {} - {}",
+        session.log_message(MessageType::INFO, format!("GoToDeclaration requested on {} at {} - {}",
```
mmahrouss: The log message is the same or declaration and definition, we should maybe set it as a parameter, or some other way, to show whether it's definition or declaration

server/src/features/declaration.rs
```diff
@@ -0,0 +1,70 @@
...
+        line: u32,
+        character: u32
+    ) -> Option<GotoDeclarationResponse> {
+        let definitions_sources = GotoUtils::get_symbols_csv(session, file_symbol, file_info, line, character);
+        let mut links = vec![];
+        for xml_result in definitions_sources.iter() {
+            links.extend(GotoUtils::goto_source_to_location(session, xml_result));
+        }
+        Some(GotoDeclarationResponse::Link(links))
+    }
```
mmahrouss: There is a lot of code dup, we could also extract it into GoToUtils

So in GoToUtils: 
```rust

    pub fn get_links<F1, F2, T>(
        session: &mut SessionInfo,
        file_symbol: &Rc<RefCell<Symbol>>,
        file_info: &Rc<RefCell<FileInfo>>,
        line: u32,
        character: u32,
        go_to_req: GotoRequest,
        get_sources: F1,
        response_constructor: F2,
    ) -> Option<T>
    where
        F1: FnOnce(&mut SessionInfo, GotoRequest, &Rc<RefCell<Symbol>>, &Rc<RefCell<FileInfo>>, u32, u32) -> Vec<GotoSource>,
        F2: Fn(Vec<LocationLink>) -> T,
    {
        let definitions_sources = get_sources(session, go_to_req, file_symbol, file_info, line, character);
        if !definitions_sources.is_empty() {
            Some(
                response_constructor(
                    definitions_sources.iter().flat_map(|r| GotoUtils::goto_source_to_location(session, r)).collect()
                )
            )
        } else {
            None
        }
    }
```` 

Then in get_location_csv: 
```rust

    pub fn _get_location_csv(session: &mut SessionInfo,
        file_symbol: &Rc<RefCell<Symbol>>,
        file_info: &Rc<RefCell<FileInfo>>,
        line: u32,
        character: u32
    ) -> Option<GotoDeclarationResponse> {
        GotoUtils::get_links(
            session,
            file_symbol,
            file_info,
            line,
            character,
            GotoRequest::Declaration,
            |session, _goto_req, file_symbol, file_info, line, character|
                GotoUtils::get_symbols_csv(session, file_symbol, file_info, line, character),
            GotoDeclarationResponse::Link)
    }
``` 

Also, we could remove the _prefix, this also works for definition, so we have much less code duplication, and also return None when there are no results

server/src/core/odoo.rs
mmahrouss: Also, those methods do not need to  be pub

server/src/core/symbols/symbol.rs
```diff
@@ -2408,6 +2444,18 @@ impl Symbol {
...
+    /* Return an iterator on all symbols of self and their sub-symbols recursively. only symbols in symbols and module_symbols will
+    * be returned.
+    */
+    pub fn all_symbols_recursive(&self) -> impl Iterator<Item= Rc<RefCell<Symbol>>> + use<> {
+        let mut iter: Vec<Rc<RefCell<Symbol>>> = Vec::new();
+        for symbol in self.all_symbols() {
+            iter.push(symbol.clone());
+            iter.extend(symbol.borrow().all_symbols_recursive());
+        }
+        iter.into_iter()
```
mmahrouss: I think it would be nice to not call the Vec iter? perhaps result?

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+            }
+            if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05056, &[model_name.as_str()]) {
+                diagnostics.push(Diagnostic {
+                    range: Range { start: Position::new(0, 0), end: Position::new(max_range as u32, 0) },
+                    ..diagnostic.clone()
+                });
+            }
+            csv_sym.set_build_status(BuildSteps::VALIDATION, BuildStatus::DONE);
+            file_info.borrow_mut().replace_diagnostics(BuildSteps::VALIDATION, diagnostics);
+            file_info.borrow_mut().publish_diagnostics(session);
```
mmahrouss: Maybe it would be nice to isolate this into an associated function, `finalize()`, since it is used ~3 times?

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+        let Some(csv_module) = &csv_module else {
+            csv_sym.set_build_status(BuildSteps::VALIDATION, BuildStatus::DONE);
+            return;
+        };
+        let model_main_sym = model.borrow().get_main_symbols(session, Some(csv_module.clone()));
+        let Some(model_main_sym) = model_main_sym.get(0) else { return;};
+        let mut header_is_xml = vec![];
+        {
+            if rdr.has_headers() {
+                if let Ok(header) = rdr.headers() {
```
mmahrouss:
```suggestion
            if rdr.has_headers() && let Ok(header) = rdr.headers() {
```
We can save one nesting level ;)

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+                    range: Range { start: Position::new(0, 0), end: Position::new(max_range as u32, 0) },
+                    ..diagnostic.clone()
+                });
+            }
+            csv_sym.set_build_status(BuildSteps::VALIDATION, BuildStatus::DONE);
+            file_info.borrow_mut().replace_diagnostics(BuildSteps::VALIDATION, diagnostics);
+            file_info.borrow_mut().publish_diagnostics(session);
+            return;
+        };
+        drop(csv);
```
mmahrouss: calls to `std::mem::drop` with a reference instead of an owned value does nothing
`#[warn(dropping_references)]` on by default

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+        let model_main_sym = model.borrow().get_main_symbols(session, Some(csv_module.clone()));
+        let Some(model_main_sym) = model_main_sym.get(0) else { return;};
+        let mut header_is_xml = vec![];
+        {
+            if rdr.has_headers() {
+                if let Ok(header) = rdr.headers() {
+                    let mut start = 0;
+                    let mut end = 0;
+                    for h in header.iter() {
+                        end = start + h.len() as u32;
```
mmahrouss: value assigned to `end` is never read
maybe it is overwritten before being read?
`#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default
```suggestion
                    for h in header.iter() {
                        let end = start + h.len() as u32;
```

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+                    }
+                }
+            }
+        }
+        csv_sym.set_build_status(BuildSteps::VALIDATION, BuildStatus::DONE);
+        file_info.borrow_mut().replace_diagnostics(BuildSteps::VALIDATION, diagnostics);
+        file_info.borrow_mut().publish_diagnostics(session);
+    }
+
+    fn validate_record(&self, session: &mut SessionInfo, csv_module: &Rc<RefCell<Symbol>>, model_name: OYarn, headers_is_xml: &Vec<bool>, record: &StringRecord, diagnostics: &mut Vec<Diagnostic>) {
```
mmahrouss: unused variable: `model_name`
`#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default

```suggestion
                        self.validate_record(session, csv_module, &header_is_xml, &result, &mut diagnostics);
                    }
                }
            }
        }
        csv_sym.set_build_status(BuildSteps::VALIDATION, BuildStatus::DONE);
        file_info.borrow_mut().replace_diagnostics(BuildSteps::VALIDATION, diagnostics);
        file_info.borrow_mut().publish_diagnostics(session);
    }

    fn validate_record(&self, session: &mut SessionInfo, csv_module: &Rc<RefCell<Symbol>>, headers_is_xml: &Vec<bool>, record: &StringRecord, diagnostics: &mut Vec<Diagnostic>) {
```

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+        if record.position().is_none() {
+            return;
+        }
+        let mut start = record.position().unwrap().byte();
+        let mut last_end = 0;
```
mmahrouss: Unused value
```suggestion
```

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+                                    }
+                                }
+                            }
+                            header_is_xml.push(is_relational);
+                        }
+                        start = end + 1;
+                    }
+                }
+            }
+            if !csv.headers.is_empty() && csv.headers[0] == "id" {
```
mmahrouss:
```suggestion
            if csv.headers.contains(&Sy!("id")){
```

server/src/core/csv_arch_builder.rs
mmahrouss:
```suggestion
            if csv.headers.contains(&Sy!("id")) {
```

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+                for result in rdr.records() {
+                    if let Ok(result) = result {
+                        self.validate_record(session, csv_module, model_name.clone(), &header_is_xml, &result, &mut diagnostics);
+                    }
+                }
```
mmahrouss:
```suggestion
                for result in rdr.records().flatten() {
                    self.validate_record(session, csv_module, model_name.clone(), &header_is_xml, &result, &mut diagnostics);
                }
```

Or maybe this is clearer that we are checking if it is okay

```suggestion
                for result in rdr.records().filter_map(Result::ok) {
                    self.validate_record(session, csv_module, model_name.clone(), &header_is_xml, &result, &mut diagnostics);
                }
```

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+                            let mut is_relational = false;
+                            for sym in member_sym.0.iter() {
+                                if sym.borrow().is_specific_field(session, &["Many2one", "One2many", "Many2many"]) {
+                                    if header_elts.len() == 2 && header_elts[1] == "id" {
+                                        //on relational fields, the header must contain :id to search for xml_id. 
+                                        //Else, it will use _rec_name and _rec_name_search to find the right record, which we are not validating
+                                        is_relational = true;
+                                        break;
+                                    }
+                                }
```
mmahrouss: Nested ifs can be one if with ands
```suggestion
                                if sym.borrow().is_specific_field(session, &["Many2one", "One2many", "Many2many"]) 
                                    && header_elts.len() == 2
                                    && header_elts[1] == "id"
                                {
                                    //on relational fields, the header must contain :id to search for xml_id. 
                                    //Else, it will use _rec_name and _rec_name_search to find the right record, which we are not validating
                                    is_relational = true;
                                    break;
                                }
```

server/src/core/csv_validation.rs
mmahrouss: I think `header_is_xml` should be exactly the same size as headers. and only when headers exist
```suggestion
        {
            if rdr.has_headers() {
                if let Ok(header) = rdr.headers() {
                    let mut start = 0;
                    let mut end = 0;
                    for h in header.iter() {
                        end = start + h.len() as u32;
                        let mut header_is_xml = vec![false; header.len()];
```
and same with `validate_record` should also be under the check of has headers

server/src/core/csv_validation.rs
mmahrouss: we can move this under the has header check, then also remove the extra scope to be: 
```rust
if !rdr.has_headers(){
   Self:finalize()
  return;
}
let Ok(header) = rdr.headers()  else {
   Self:finalize()
  return;
} 
``` 
then we continue processing the headers and records. 
We will have much less nesting and more readability

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn validate_record(&self, session: &mut SessionInfo, csv_module: &Rc<RefCell<Symbol>>, model_name: OYarn, headers_is_xml: &Vec<bool>, record: &StringRecord, diagnostics: &mut Vec<Diagnostic>) {
+        if record.position().is_none() {
+            return;
+        }
+        let mut start = record.position().unwrap().byte();
+        let mut last_end = 0;
+        for (idx, field) in record.iter().enumerate() {
+            let end = start + field.len() as u64;
+            let Some(should_be_xml_id) = headers_is_xml.get(idx) else { break;};
```
mmahrouss, line 109: we should be able to zip records, and headers_is_xml, and have them in the interator, they must be the same length.

server/src/core/evaluation.rs
```diff
@@ -709,51 +716,107 @@ impl Evaluation {
...
             ExprOrIdent::Expr(Expr::Call(expr)) => {
+                // Check argument expressions for references and diagnostics
+                if is_in_validation || odoo.evaluation_search.is_some() {
+                    for param in expr.arguments.args.iter().chain(expr.arguments.keywords.iter().map(|k| &k.value)) {
+                        let (_, diags) = Evaluation::eval_from_ast(session, &param, parent.clone(), max_infer, false, required_dependencies);
+                        diagnostics.extend(diags);
+                    }
+                }
                 let (base_evals, diags) = Evaluation::eval_from_ast(session, &expr.func, parent.clone(), max_infer, false, required_dependencies);
                 diagnostics.extend(diags);
```
mmahrouss: consider checking if in validation before adding to diagnostic
```suggestion
                if is_in_validation { 
                    diagnostics.extend(diags);
                 }
```

server/src/core/python_arch_eval.rs
```diff
@@ -314,12 +314,26 @@ impl PythonArchEval {
                 expr_slice.lower.as_ref().map(|lower_expr| self.visit_expr(session, &lower_expr));
             },
             // Expressions that cannot contained a named expressions are not traversed
-            Expr::Lambda(_todo_lambda_expr) => {
+            Expr::Lambda(lambda_expr) => {
                 // Lambdas can have named expressions, but it is not a common use
                 // Like lambda vals: vals[(x := 0): x + 3]
                 // However x is only in scope in the lambda expression only
                 // It needs adding a new function, ast_indexes, then add the variable inside
                 // I deem it currently unnecessary
```
mmahrouss: I guess we can drop my comment oops 🗡️

server/src/features/csv_ast_utils.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub fn get_symbols(session: &mut SessionInfo, file_symbol: &Rc<RefCell<Symbol>>, csv_reader: &mut Reader<&[u8]>, model_name: &OYarn, offset: usize) -> Vec<GotoSource> {
+        let mut results = vec![];
+        let module = file_symbol.borrow().find_module();
+        let Some(model) = session.sync_odoo.models.get(model_name).cloned() else {return vec![];};
+        let model_syms = model.borrow().get_main_symbols(session, module.clone());
+        let Some(main_symbol) = model_syms.first().cloned() else {return results;};
+        drop(model_syms);
+        let mut headers = vec![];
+        if csv_reader.has_headers() {
+            if let Ok(header) = csv_reader.headers() {
```
mmahrouss: we can reduce nesting by early returns 
```suggestion
        if !csv_reader.has_headers() {
            return vec![];
        }
        let Ok(header) = csv_reader.headers() else { return vec![]; };
        let mut results = vec![];
        ...
        return results
```

server/src/features/csv_ast_utils.rs
```diff
@@ -0,0 +1,148 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    h_start = end + 1;
+                }
+            }
+        }
+        if !headers.is_empty() && headers[0] == "id" {
```
mmahrouss:
```suggestion
        if headers.contains(&oyarn!("id")) {
```

server/src/features/csv_ast_utils.rs
```diff
@@ -0,0 +1,148 @@
...
+            for result in csv_reader.records() {
+                if let Ok(record) = &result {
+                    CsvAstUtils::get_symbols_in_record(session, offset, &headers, &record, &mut results, &file_symbol, model_name.clone(), &main_symbol, module.clone());
+                }
+            }
```
mmahrouss: we can be concise 
```suggestion
            for result in csv_reader.records().filter_map(Result::ok) {
                CsvAstUtils::get_symbols_in_record(session, offset, &headers, &result, &mut results, &file_symbol, model_name.clone(), &main_symbol, module.clone());
            }
```

server/src/features/csv_ast_utils.rs
```diff
@@ -0,0 +1,148 @@
...
+                                    &mut vec![]
+                                ).iter().map(|odoo_data| GotoSource {
+                                    source: GotoSourceType::OdooData(odoo_data.clone()),
+                                    origin_selection_range: None,
+                                }));
+                            }
+                        }
+                    }
+                }
+            }
```
mmahrouss, line 133: We could flatten the for loops a bit by pre-calculating the related models then processing it. It is just a suggestion

```suggestion
                // 1. find relational field in current model
                let field_syms = main_symbol.borrow().get_member_symbol(session, &S!(field_name), module.clone(), false, true, false, true, false).0;
                let relational_models = field_syms
                .iter()
                .filter_map(|field_sym| {
                    // 1.1 Filter out relational fields
                    if field_sym.borrow().is_specific_field(session, &["Many2one", "One2many", "Many2many"])
                        && field_sym.borrow().typ() == SymType::VARIABLE
                    {
                        // 1.2 Get associated relational model
                        let relational_models = field_sym.borrow().as_variable().get_relational_model(session, module.clone());
                        Some(relational_models)
                    } else {
                        None
                    }
                })
                .flatten()
                .collect::<Vec<_>>(); // 1.3 Flatten out to get all related models main symbols
                for related_model in relational_models.iter() {
                    // 3. find related field in related model
                    let related_syms = related_model.borrow().get_member_symbol(session, &S!(*relational_field), module.clone(), false, true, false, true, false).0;
                    // 4. push result in results
                    if !related_syms.is_empty() {
                        results.extend(SyncOdoo::get_xml_ids(session,
                            &csv_symbol,
                            field_data.as_str(),
                            &std::ops::Range::default(), //we don't care about range as it's used only for diagnostic
                            &mut vec![]
                        ).iter().map(|odoo_data| GotoSource {
                            source: GotoSourceType::OdooData(odoo_data.clone()),
                            origin_selection_range: None,
                        }));
                    }
                }
```

server/src/features/references.rs
```diff
@@ -1,70 +1,551 @@
...
+                                    let data = file_info.borrow().file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
+                                    let document = roxmltree::Document::parse(&data);
+                                    if let Ok(document) = document {
+                                        let root = document.root_element();
+                                        locations.extend(XmlAstReferenceVisitor::search_target(session, &file_s, root, &ReferenceTarget::String(full_xml_id)));
+                                    }
+                                },
+                                SymType::CSV_FILE => {
+                                    //let model_name_pb = PathBuf::from(&file_s.borrow().paths()[0]);
+                                    //let model_name = Sy!(model_name_pb.file_stem().unwrap().to_str().unwrap().to_string());
```
mmahrouss: Orphaned comment

mmahrouss (commented):
it is mostly style comments, feel free to ignore some of them:) 
I did not manage to read everything

server/src/features/csv_ast_utils.rs
```diff
@@ -0,0 +1,146 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            h_start = end + 1;
+        }
+        if headers.contains(&oyarn!("id")) {
+            for record in csv_reader.records().filter_map(Result::ok) {
```
cammarosano, line 62: Wicked! ❤️

server/src/features/references.rs
```diff
@@ -1,70 +1,550 @@
...
+                    let mut to_check: PtrWeakHashSet<Weak<RefCell<Symbol>>> = PtrWeakHashSet::new();
+
+                    to_check.insert(file_symbol.clone());
+
+                    if let Some(target_file) = target_symbol.borrow().get_file().and_then(|f| f.upgrade()) {
+                        //take arch and arch_eval dependents
+                        if !target_file.borrow().dependents().is_empty() { // file could be out of workspace
+                            for dep in target_file.borrow().dependents().iter().take(2) {
+                                //dep.len()-1 here is to take only dependencies that are not validation (arch and arch_eval for arch, arch_eval for arch_eval)
+                                for dep in dep.iter().take(dep.len()) {
```
cammarosano: comment says `dep.len() -1`, code does `dep.len()`.
Anyway, this `take` in `dep.iter().take(dep.len())` is a no-op (you're just taking everything), might as well just call `dep.iter()` (without `take`)

server/src/features/references.rs
```diff
@@ -1,70 +1,550 @@
...
+                            for data in module.borrow().as_module_package().data_symbols.values().cloned() {
+                                files_to_process.insert(data);
+                            }
+                        }
+                    }
+                    //add python dependencies
+                    if !current_module.borrow().dependents().is_empty() { // file could be out of workspace
+                        for dep in current_module.borrow().dependents().iter().take(2) {
+                            //dep.len()-1 here is to take only dependencies that are not validation (arch and arch_eval for arch, arch_eval for arch_eval)
+                            for dep in dep.iter().take(dep.len()) {
```
cammarosano: same as above (comment does not match code, `.take(dep.len())` is no-op

server/src/core/csv_validation.rs
```diff
@@ -0,0 +1,144 @@
...
+                        });
+                    }
+                    continue;
+                };
+                let Some(module) = module_symbol.upgrade() else {continue};
+                let module_borrow = module.borrow();
+                if module_borrow.as_module_package().xml_id_locations.get(&Sy!(id_split.last().unwrap().to_string())).is_none() {
+                    if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05001, &[]) {
+                        diagnostics.push(Diagnostic {
+                            range: Range { start: Position::new(start as u32, 0), end: Position::new(end as u32+1, 0) },
```
cammarosano: as we discussed live, the comma ("," ) after the value is getting included in the squiggle range, you should probably remove this `+1`.

server/src/features/references_csv.rs
```diff
@@ -0,0 +1,133 @@
...
+        locations
+    }
+
+    pub fn remove_quotes(s: &str) -> String {
+        if s.starts_with('"') && s.ends_with('"') && s.len() >= 2 {
+            s[1..s.len()-1].to_string()
+        } else {
+            s.to_string()
+        }
+    }
```
cammarosano: This function is duplicated! Can't you  just use the one from csv_ast_utils in this file?

server/src/core/evaluation.rs
```diff
@@ -709,53 +716,111 @@ impl Evaluation {
...
                 }
-                evals.push(Evaluation::new_dict(odoo, values, expr.range));
+                evals.push(Evaluation::new_dict(session.sync_odoo, Some(values), expr.range));
             },
             ExprOrIdent::Expr(Expr::Call(expr)) => {
+                // Check argument expressions for references and diagnostics
+                if is_in_validation || odoo.evaluation_search.is_some() {
+                    for param in expr.arguments.args.iter().chain(expr.arguments.keywords.iter().map(|k| &k.value)) {
+                        let (_, diags) = Evaluation::eval_from_ast(session, &param, parent.clone(), max_infer, false, required_dependencies);
+                        diagnostics.extend(diags);
```
cammarosano, line 817: Shouldn't this be guarded by `if is_in_validation`, like in other places in this function? Otherwise you're (potentially) doing unnecessary allocation for a vector you're not  reading afterwards.
```suggestion
              if is_in_validation {
                  diagnostics.extend(diags);
              }
              
```

server/src/core/evaluation.rs
```diff
@@ -1266,11 +1324,11 @@ impl Evaluation {
...
             },
             ExprOrIdent::Expr(Expr::BinOp(operator)) => {
-                match operator.op {
-                    Operator::Add => {
-
-                    },
-                    _ => {}
+                if is_in_validation || odoo.evaluation_search.is_some() {
+                    let (_, diags) = Evaluation::eval_from_ast(session, &operator.left, parent.clone(), max_infer, false, required_dependencies);
+                    diagnostics.extend(diags);
```
cammarosano, line 1329: same as above. Probably should be wrapped in a `if _is_in_validation` block

server/src/core/evaluation.rs
```diff
@@ -1266,11 +1324,11 @@ impl Evaluation {
...
+                    let (_, diags) = Evaluation::eval_from_ast(session, &operator.right, parent.clone(), max_infer, false, required_dependencies);
+                    diagnostics.extend(diags);
```
cammarosano, line 1331: same!

server/tests/test_references.rs
cammarosano, line 1: 😍

cammarosano (approved):
Very nice super cool amazing job!. Killer feature!

server/src/core/evaluation.rs
fda-odoo: It is already :)

server/src/features/references_csv.rs
fda-odoo: You're right, I forgot to remove previous one when I moved it to CsvAstUtils

server/src/core/evaluation.rs
fda-odoo: It is already in is_in_validation block too

server/src/core/evaluation.rs
cammarosano: If I got this right, It's in a `is_in_validation` **OR** `odoo.evaluation_search.is_some()` block.
So if you're just searching for references (thus, not validating), you're populating the `diagnostics` vector needlessly.

PR #551, opened 2026-02-25
[FIX] fix gotodef on model xml_ids

mmahrouss:
First, go to definition on xml_ids from python code was panicking. In this PR, we get the file of the xml_Id symbol.
Then we make sure to get the correct range of the symbol, not the file.

### DIFF
--- a/server/src/features/definition.rs
+++ b/server/src/features/definition.rs
@@ -131,17 +131,21 @@ impl DefinitionFeature {
         let mut xml_found = false;
         let xml_ids = SyncOdoo::get_xml_ids(session, file_symbol, value.as_str(), &std::ops::Range{start: 0, end: 0}, &mut vec![]);
         for xml_id in xml_ids {
-            let file = xml_id.get_file_symbol();
-            if let Some(file) = file {
-                if let Some(file) = file.upgrade() {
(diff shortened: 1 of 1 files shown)


server/src/features/definition.rs
```diff
@@ -131,17 +131,21 @@ impl DefinitionFeature {
...
-                    let range = session.sync_odoo.get_file_mgr().borrow().std_range_to_range(session, &file.borrow().paths()[0], &xml_id.get_range());
-                    xml_found = true;
-                    links.push(LocationLink {
-                        origin_selection_range: eval.range.map(|r| session.sync_odoo.get_file_mgr().borrow().text_range_to_range(session, file_symbol.borrow().paths().first().as_ref().unwrap(), &r)),
-                        target_uri: FileMgr::pathname2uri(&file.borrow().paths()[0]),
-                        target_range: range,
-                        target_selection_range: range });
-                }
+            let maybe_file = xml_id
+                .get_file_symbol()
```
fda-odoo, line 135: if get_file_symbol is not returning a file_symbol, then we should change the name, or the way it works. I got the same issue in my branch, I will include this fix in it

fda-odoo (changes requested):

PR #557, opened 2026-02-27
Process config files in parse (cli) mode, and allow selecting config profile through arguments

mmahrouss:
Still need to figure out how to test it

### DIFF
--- a/server/src/cli_backend.rs
+++ b/server/src/cli_backend.rs
@@ -3,68 +3,191 @@ use lsp_types::notification::{LogMessage, Notification, PublishDiagnostics};
 use lsp_types::{LogMessageParams, PublishDiagnosticsParams, Uri};
 use tracing::{error, info};
 
-use crate::core::config::ConfigEntry;
-use crate::threads::SessionInfo;
-use crate::utils::{get_python_command, PathSanitizer};
-use crate::args::Cli;
-use std::io::Write;
-use std::path::PathBuf;
-use std::fs::{self, File};
-use std::str::FromStr;
-use serde_json::json;
-use crate::core::{config::{DiagMissingImportsMode}, odoo::SyncOdoo};
 use crate::S;
+use crate::args::Cli;
+use crate::core::config::{ConfigEntry, get_configuration};
+use crate::core::odoo::SyncOdoo;
+use crate::threads::SessionInfo;
+use crate::utils::{PathSanitizer, is_addon_path, is_odoo_path, is_python_path};
+use serde_json::json;
+use std::collections::HashMap;
+use std::fs::{self, File};
+use std::io::Write;
+use std::path::Path;
+use std::str::FromStr;
 
+    fn canonicalize_and_validate(
+        path: &str,
+        is_valid_str: Option<fn(&String) -> bool>,
+        is_valid_pb: Option<fn(&Path) -> bool>,
+        error_msg: &str,
+    ) -> Option<String> {
+        match fs::canonicalize(path) {
+            Ok(pb) => {
+                let sanitized = pb.sanitize();
+                let valid = match (is_valid_str, is_valid_pb) {
+                    (Some(is_valid_str), Some(is_valid_pb)) => {
+                        is_valid_str(&sanitized) && is_valid_pb(&pb)
+                    }
+                    (Some(is_valid_str), None) => is_valid_str(&sanitized),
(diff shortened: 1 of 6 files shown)


server/src/core/odoo.rs
```diff
@@ -159,6 +161,7 @@ impl SyncOdoo {
         session.sync_odoo.version_micro = 0;
         session.sync_odoo.full_version = "0.0.0".to_string();
         session.sync_odoo.config = ConfigEntry::new();
+        session.sync_odoo.selected_config = None;
```
fda-odoo: so now we have `selected_config` and `selected_config_profile` on sync_odoo. I think that's confusing. Maybe rename `selected_config` by something like `cli_selected_config?`
Maybe the same for `config_path` ?

fda-odoo (changes requested):

server/src/core/odoo.rs
mmahrouss: Hmmm I think `selected_config_profile` and `selected_config`  should be the same, I will update it, and will also check `config_path` too

server/src/core/odoo.rs
```diff
@@ -1446,12 +1446,23 @@ impl Odoo {
                 return;
             }
         };
-        let selected_config = match maybe_selected_config {
-            None => default_profile_name(),
-            Some(c) if c == "" => default_profile_name(),
-            Some(config) => config,
+        let selected_config = match session.sync_odoo.selected_config {
+            Some(ref current) => {
+                info!("Using selected configuration from cli arguments ({})", current);
```
fda-odoo: technically, selected_config on sync_odoo is not a data coming from CLI. I would move this log to the cli_backend as it is there that it is really set.
cli_backend : "CLI selected config : xxx"
and here: "final selected config: xxx"

fda-odoo (approved):

server/src/core/odoo.rs
fda-odoo: I would add that CLI could log every arguments indeed used in the configuration, not only selected_config

mmahrouss: > technically, selected_config on sync_odoo is not a data coming from CLI. I would move this log to the cli_backend as it is there that it is really set. cli_backend : "CLI selected config : xxx" and here: "final selected config: xxx"

But here in this case, if you are using Zed for example, and add `--selected-config=config`, then this value if from the CLI arguments that Zed passes when running OdooLS, from the user defined binary arguments, and in that case it takes precedence.

fda-odoo: Ok I was wrong with the location in my comment. I meant `set_config_cli_args` that is in server.rs

fda-odoo: AND cli_backend

mmahrouss: Well, I am confused :D

`set_config_cli_args ` in server.rs:  "CLI selected config : xxx"

and odoo.rs/cli_backend.rs: "final selected config: xxx"? 


> I would add that CLI could log every arguments indeed used in the configuration, not only selected_config

Also this, we already log the final config in `SyncOdoo::init`

mmahrouss: I updated the logs in an extra commit :)

server/src/cli_backend.rs
```diff
@@ -3,68 +3,211 @@ use lsp_types::notification::{LogMessage, Notification, PublishDiagnostics};
...
+            Some(selected_config) => {
+                info!("CLI selected config profile: {:?}", selected_config);
+                selected_config
+            },
+            None => {
+                info!("No CLI selected config profile, using default: {:?}", DEFAULT_PROFILE_NAME);
+                DEFAULT_PROFILE_NAME.to_string()
+            }
+        };
+        info!("Selected config profile ({})", selected_config);
```
fda-odoo: why do you log it a second time here?

fda-odoo (approved):

server/src/cli_backend.rs
mmahrouss: For consistency with the behavior in Odoo.rs? 
We log what we get, and then the final one.

fda-odoo: Ok, I see the idea, but I don't think we should try to have the same behaviour when we do something else. It could be just confusing in the end to have multiple logs that say the same thing.
Here it's useless as no operation is done between the last log and this one, so I would remove it :)

mmahrouss: Alright done:)

PR #559, opened 2026-03-03
[IMP] Handle lambdas and other missing expressions

mmahrouss:


server/src/core/evaluation.rs
```diff
@@ -707,51 +715,95 @@ impl Evaluation {
...
+                let mut all_values = true;
                 let mut values: Vec<ruff_python_ast::Expr> = Vec::new();
                 for e in expr.elts.iter() {
-                    if e.is_literal_expr() {
+                    if is_in_validation || session.sync_odoo.evaluation_search.is_some() {
+                        let (_, diags) = Evaluation::eval_from_ast(session, e, parent.clone(), max_infer, false, required_dependencies);
+                        if is_in_validation {
+                            diagnostics.extend(diags);
+                        }
+                    }
```
fda-odoo, line 727: I don't know if we would gain a lot with that, but maybe we could bring back the `break` if not in validation?

```rust
if is_in_validation {
    //
} else if !all_values {
    break;
}
```

server/src/core/evaluation.rs
```diff
@@ -1354,8 +1400,161 @@ impl Evaluation {
...
+            ExprOrIdent::Expr(Expr::Lambda(lambda_expr)) => {
+                let variable: Option<Rc<RefCell<Symbol>>> = parent.borrow().get_positioned_symbol(&Sy!("<lambda>"), &lambda_expr.range);
+                if let Some(lambda_fn_sym) = variable {
+                    if is_in_validation || odoo.evaluation_search.is_some() {
+                        if is_in_validation {
+                            lambda_fn_sym.borrow_mut().set_build_status(BuildSteps::VALIDATION, BuildStatus::IN_PROGRESS);
+                        }
+                        let (_, diags) = Evaluation::eval_from_ast(session, &lambda_expr.body, lambda_fn_sym.clone(), &lambda_expr.body.range().start(), false, required_dependencies);
+                        if is_in_validation {
+                            lambda_fn_sym.borrow_mut().set_build_status(BuildSteps::VALIDATION, BuildStatus::IN_PROGRESS);
```
fda-odoo: DONE ?

server/src/core/evaluation.rs
```diff
@@ -1354,8 +1400,161 @@ impl Evaluation {
...
+            },
+            ExprOrIdent::Expr(Expr::YieldFrom(yield_from_expr)) =>{
+                if is_in_validation || odoo.evaluation_search.is_some() {
+                    let (_, diags) = Evaluation::eval_from_ast(session, &yield_from_expr.value, parent.clone(), max_infer, false, required_dependencies);
+                        if is_in_validation {
+                            diagnostics.extend(diags);
+                        }
+                }},
+            ExprOrIdent::Expr(Expr::Await(await_expr)) =>{
+                let (evaluations, diags) = Evaluation::eval_from_ast(session, &await_expr.value, parent.clone(), max_infer, false, required_dependencies);
```
fda-odoo, line 1498: why don't you check for is_in_validation here?

server/src/core/python_arch_builder.rs
```diff
@@ -456,12 +456,16 @@ impl PythonArchBuilder {
...
             },
             // Expressions that cannot contained a named expressions are not traversed
-            Expr::Lambda(_todo_lambda_expr) => {
-                // Lambdas can have named expressions, but it is not a common use
-                // Like lambda vals: vals[(x := 0): x + 3]
-                // However x is only in scope in the lambda expression only
-                // It needs adding a new function, ast_indexes, then add the variable inside
-                // I deem it currently unnecessary
+            Expr::Lambda(lambda_expr) => {
+                let sym = self.sym_stack.last().unwrap().borrow_mut().add_new_function(
```
fda-odoo, line 460: this is unfortunately not perfectly right. As you create a 'lambda' function in the tree, it makes this symbol available later, so we could validate
```python
a = lambda x: 5
print(lambda)
```
Even if syntax is ofc invalid, it means that this lambda function is available in the scope. And we have nothing that prevent using a reserved keyword in the scope. But I agree it will work in normal cases.

fda-odoo (changes requested):

server/src/core/evaluation.rs
mmahrouss: Because I am actually using the evaluation in an await expression, it can be used somewhere.

server/src/core/evaluation.rs
mmahrouss: Yes, I could do that

server/src/core/evaluation.rs
mmahrouss: Yes oops

server/src/core/python_arch_builder.rs
mmahrouss: but it is `"<lambda>"`, so it would not be present in any python code, no ?

PR #561, opened 2026-03-10
[IMP] Add XML model records to models

mmahrouss:
Depends on the fix to go to def on XMl panicking or going to wrong file https://github.com/odoo/odoo-ls/pull/551/changes#r2953025503

### DIFF
--- a/server/src/core/xml_validation.rs
+++ b/server/src/core/xml_validation.rs
@@ -43,7 +43,7 @@ impl XmlValidator {
                 self.validate_xml_id(session, &module, xml_id, &mut diagnostics, &mut dependencies, &mut model_dependencies, &mut missing_model_dependencies);
             }
         }
-        for dep in dependencies.iter_mut() {
+        for dep in dependencies.iter_mut().filter(|dep_sym| !Rc::ptr_eq(*dep_sym, &self.xml_symbol)) {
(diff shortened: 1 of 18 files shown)


server/src/core/xml_validation.rs
```diff
@@ -123,11 +179,29 @@ impl XmlValidator {
...
                     if has_required && !has_default {
-                        mandatory_fields.push(field_name.to_string());
+                        mandatory_fields.insert(field_name.to_string());
                     }
                 }
             }
         }
+        // Also compute mandatory fields from XML-defined ir.model.fields records
+        for (field_name, field_records) in xml_field_records.iter() {
+            if mandatory_fields.contains(field_name) {
```
mmahrouss, line 187: What if an XML field is just added to make a field not required?

fda-odoo:
can you rebase this one too? @mmahrouss

mmahrouss:
@fda-odoo rebased

fda-odoo (changes requested):
We have to discuss this PR, because I think a refactoring is needed at this point.
Model class has been designed to only keep reference to symbols that build the model. Here it holds copies of OdooDataRecord.
Maybe we have to update the structure so that OdooDataRecord are part of the Symbol tree.
get_member is wrong too, as it is not browsing xml_symbols added to Model, same for all_members

mmahrouss:
To be addressed at https://github.com/odoo/odoo-ls/pull/600 which is based on a later revision

PR #566, opened 2026-03-18
[FIX] Fix first argument on static methods

mmahrouss:
Before first arg was always linked the parent class. But this is not true for static methods. Make sure we check for those here


fda-odoo (approved):

fda-odoo:
@mmahrouss tests are failing?

PR #571, opened 2026-03-24
Validate assets structure in manifests

fda-odoo:


server/src/core/symbols/module_symbol.rs
```diff
@@ -254,6 +256,132 @@ impl ModuleSymbol {
                                         }
                                     }
                                 }
+                            } else if key_str == "assets" {
```
mmahrouss, line 200: I feel that `_load_manifest` has become too long to be one function, perhaps we can create associated functoons `load_assets`, `load_data`, ... etc

server/src/core/diagnostic_codes_list.rs
```diff
@@ -235,6 +235,38 @@ OLS04011, DiagnosticSetting::Error, "Do not use dict unpacking to build your man
 */
 OLS04012, DiagnosticSetting::Error, "Module dependency: module {0} depends on current module",
 /**
+* The assets value should be a dictionnary
+*/
+OLS04013, DiagnosticSetting::Error, "The assets value should be a dictionnary",
+/**
+* 'OR' Expression is not allowed in assets definition.
+*/
+OLS04014, DiagnosticSetting::Error, "'OR' Expression is not allowed in assets definition.",
```
mmahrouss, line 244: What is an 'OR' Expression?

server/src/core/diagnostic_codes_list.rs
```diff
@@ -235,6 +235,38 @@ OLS04011, DiagnosticSetting::Error, "Do not use dict unpacking to build your man
...
+*/
+OLS04016, DiagnosticSetting::Error, "Bundle content should be expressed with a list of strings",
+/**
+* An element of a bundle should be a string or a tuple of ('action': string, 'path': string)
+*/
+OLS04017, DiagnosticSetting::Error, "An element of a bundle should be a string or a tuple of ('action': string, 'path': string)",
+/**
+* Tuple element should be a string
+*/
+OLS04018, DiagnosticSetting::Error, "Tuple element should be a string",
```
mmahrouss: Maybe specify "in assets" or something similar?

mmahrouss (approved):

server/src/core/diagnostic_codes_list.rs
fda-odoo: that's a part of the AST I guess nobody will use, but it's 
```python
("stg": 5, **another_dict)
```

server/src/core/diagnostic_codes_list.rs
fda-odoo: I'll update it, but the diagnostic can be raised in assets only anyway

server/src/core/symbols/module_symbol.rs
fda-odoo: tab tab tab tab tab tab I agree
tab tab tab tab tab tab tab I'll split that so we can have something more clear

server/src/core/diagnostic_codes_list.rs
```diff
@@ -235,6 +235,38 @@ OLS04011, DiagnosticSetting::Error, "Do not use dict unpacking to build your man
 */
 OLS04012, DiagnosticSetting::Error, "Module dependency: module {0} depends on current module",
 /**
+* The assets value should be a dictionnary
+*/
+OLS04013, DiagnosticSetting::Error, "The assets value should be a dictionnary",
```
cammarosano: typo (dictionary, 1 n only)

cammarosano (approved):
Nice!
(but you should fix the typo, considering it's a user-facing string)

PR #574, opened 2026-04-01
Fix CSV reading

cammarosano:


server/src/features/csv_ast_utils.rs
```diff
@@ -1,16 +1,69 @@
...
+    record: &'a StringRecord,
+    field_idx: usize,
+    start: usize,
+}
+
+impl<'a> CsvFieldIter<'a> {
+    pub fn new(record: &'a StringRecord, content: &'a str) -> Option<Self> {
+        let position = record.position()?;
+        let mut start = position.byte() as usize;
+        if content.as_bytes().get(start) == Some(&b'\n') {
```
cammarosano, line 23: maybe add a comment about the CRLF messing up with with the offset returned by `record.position()`

server/tests/test_get_symbol.rs
cammarosano, line 1: 😍

fda-odoo (approved):

PR #581, opened 2026-04-10
[REF] store symbols in arena

cammarosano:
## Refactor: Replace `Rc<RefCell<Symbol>>` with a slotmap-based `SymbolTable`

----------
This PR migrates from the old reference-counted symbol graph
(`Rc<RefCell<Symbol>>` + `Weak<RefCell<Symbol>>`) to a centralized arena
(`SymbolTable`) backed by [slotmap](https://docs.rs/slotmap).

## 1. Storage model: one slotmap per variant

`SymbolTable` holds **one `SlotMap` per concrete symbol variant** (`classes`,
`functions`, `variables`, `files`, `modules`, `python_packages`, `namespaces`,
`disk_dirs`, `compiled`, `xml_files`, `csv_files`, `roots`). Each variant has
its own typed key (`ClassKey`, `FunctionKey`, …). A top-level `SymbolKey` enum
wraps them and is used wherever code needs to be variant-agnostic.

Using one map per variant rather than a single `SlotMap<SymbolKey, Symbol>`
avoids paying the max-variant padding cost on every slot — the variants have
very different sizes.

Relevant files:

- `server/src/core/symbols/symbol_keys.rs`
- `server/src/core/symbols/storage.rs`

---

### Strong keys — always valid

**Every `SymbolKey` (or typed variant key) held in the tree is a strong key: the
symbol table guarantees the symbol it points to still lives in the table.**
Strong keys can be indexed directly (`table[key]`) with no `Option` unwrap. In
practice, strong keys are **parent keys** (every symbol stores its parent as a
plain key) and **children keys** (`symbols`, `module_symbols` on the variant
structs).

This guarantee is enforced by two mechanisms at the module boundary:

1. **Private slotmaps.** All `SlotMap` fields on `SymbolTable` are
   private. Only code inside the `storage` module can call
   `.insert()` / `.remove()`. Everywhere else uses the `Index<TypedKey>` trait,
   which is gived access to the symbol inside the slot, but not the slot itself.
2. **Restricted children mutation.** `symbols` / `module_symbols` on the
   variant structs are `pub(super)` — writable only from inside
   `storage`. They can be read via getters (`parent()`, `symbols()`), but
   mutation happens exclusively via `SymbolTable`'s `add_new_*` / `unload` pub
   methods.

Because no code outside `storage` can create, move, or destroy symbols, any key
reachable through `parent` or `symbols` / `module_symbols` is trustworthy — no
upgrade needed.

### Weak keys — must be upgraded

Everywhere else — caches, dependency sets, ext-symbol stores, evaluations — uses
`Weak<K>`, the arena equivalent of `std::rc::Weak`:

A `Weak<SymbolKey>` must be `upgrade`d to a `SymbolKey` against the table before
use. You can't do anything with a `Weak<SymbolKey>` except check if it's expired
or try to upgrade it.

Collections of weak keys are held in a dedicated `WeakSet<K>`. It deliberately
does not expose a plain `iter()` — the only way to walk it is
`iter_valid(&symbol_table)`, which upgrades each entry against the table and
yields only the valid ones. Stale entries are therefore cleaned lazily on
iteration, replacing the old `PtrWeakHashSet` auto-expiry.

Any key stored outside the tree (i.e. outside parent/child relations) must be
stored as a `Weak<K>` or inside a `WeakSet<K>`.

Relevant files:

- `server/src/core/symbols/symbol_keys.rs` — `Weak<K>`, `KeyValidator` trait
- `server/src/weak_collections.rs` — `WeakSet<K>`, `iter_valid`
- `server/src/core/symbols/storage.rs` — private slotmaps, `KeyValidator` impls
- `server/src/core/symbols/storage/lifecycle.rs` — the only place that
  mutates slotmaps and `pub(super)` children fields
- `server/src/core/symbols/storage/class_symbol.rs`,
  `file_symbol.rs`, … — `pub(super)` `symbols` / `module_symbols`
  declarations

---

## 3. New API

`SymbolTable` lives on `SyncOdoo` and is reached as
`session.sync_odoo.symbol_table`. A local `st!()` macro shortens this inside
functions. (this is being replaced by `session.st()` and `session.st_mut()`
convenience methods)

### Reading

SymbolTable implements `Index` for specific keys, and its the only way to access symbols from outside the storage module.

```rust
let name   = &st!()[file_key].name;
let parent = st!()[class_key].parent;
```

When the variant is not statically known, you either pattern-match on
`SymbolKey` to recover a typed key, or call one of the variant-agnostic accessor
methods on `SymbolTable` that do the pattern matching for you and return a
common field — e.g. `st!().name(key)` or `st!().is_external(key)` (like the former Symbol methods)

### Creating / removing

All creation goes through `add_new_*` methods on `SymbolTable` (`add_new_class`,
`add_new_function`, …). They insert into the right slotmap, register the new
symbol in its parent's children, inherit `is_external`, and return the **typed**
key. Removal goes through `remove_symbol` / `unload`. There is no other way to
mutate the slotmaps.

### Queries and navigation

The old `Symbol` methods and free functions are still in the `symbols.rs` module,
but now as methods or associated functions of `SymbolTable`. The former `self` parameter
is renamed `target` in most methods, since `self` is now the `SymbolTable`
instance.

### External symbols

`ext_symbols` / `decl_ext_symbols` used to live on some variant structs as two
almost-always-empty `HashMap`s. They are now centralized in a single
`ExtSymbolStore` on `SymbolTable`, which saves a significant amount of space.

Relevant files:

- `server/src/core/symbols/storage/lifecycle.rs` — `add_new_*`,
  `remove_symbol`, `unload`
- `server/src/core/symbols/symbols.rs` — query / navigation / eval
  methods
- `server/src/core/symbols/storage/ext_symbol_store.rs` — `ExtSymbolStore`
- `server/src/core/symbols/storage/dependency_mgr.rs` — `Dependencies` /
  `Buildable` traits + macro to avoid code duplication

---

### No more runtime borrow errors (for Symbols)

Now compile-time checked.

### Memory layout

A little less data, less indirection, a lot less padding.

### Ergonomics

`SymbolKey` and the typed keys are `Copy` (~12 and ~8 bytes). This is what
removes most of the noise at call sites: no more `.clone()` / `Rc::clone(&sym)`,
no more `.borrow()` / `.borrow_mut()` pairs, no more `Rc<RefCell<Symbol>>` in
signatures. Field access is just `table[key].field` instead of
`symbol.borrow().as_variable().field`.

### Type safety

Functions that previously took `Rc<RefCell<Symbol>>` and immediately asserted
`sym.typ() == SymType::CLASS` now take a `ClassKey` directly — the wrong-variant
case is ruled out at compile time. Pattern matching on `SymbolKey` (`if let
SymbolKey::Variable(v) = key`) replaces the old `sym.borrow().typ() == …` checks
and extracts a typed key in one step.

---

## 5. Other things worth noting

- **`weak_self` eliminated.**
- **`parent`** is no longer Option (and Root has no `parent`)
- **File layout.**
  ```
  core/symbols/
    symbol_keys.rs         — typed keys, SymbolKey enum, Weak<K>, KeyValidator
    storage.rs             — SymbolTable struct declaration, private
                             slotmaps, Index / KeyValidator impls
    storage/               — variant structs + everything that needs
                             pub(super) access to their children fields
      class_symbol.rs, file_symbol.rs, …  — the variant structs
      lifecycle.rs         — all symbol creation / removal / unload
                             (the only place that inserts into or
                              removes from the private slotmaps)
      symbol_mgr.rs, dependency_mgr.rs, ext_symbol_store.rs
    symbols.rs          — query / navigation / eval methods on
                             SymbolTable (no slotmap or pub(super) access)
  ```
  `impl SymbolTable` blocks are deliberately split across several files. Only
  the ones physically inside the `storage` module (notably `lifecycle.rs`) can
  touch the private slotmaps or the `pub(super)` `symbols` / `module_symbols`
  fields on the variants. Everything in `symbols.rs` or outside
  `core::symbols` is read-only with respect to the tree structure, which is what
  makes the strong-key guarantee from §2 enforceable by the compiler.

---

## 6. Suggested review order

1. `symbol_keys.rs` — typed keys, `SymbolKey`, `Weak<K>`, `KeyValidator`.
   Establishes the vocabulary used everywhere else.
2. `storage.rs` — `SymbolTable` struct declaration, private slotmaps, `Index` /
   `IndexMut` / `KeyValidator` impls. This is where the "strong key" invariant
   starts.
3. `storage/class_symbol.rs`, `file_symbol.rs`, `function_symbol.rs`, … — the
   variant structs, to see which fields are `pub(super)` and how children /
   parent relations are stored.
4. `storage/lifecycle.rs` — the only place that inserts into or removes from the
   slotmaps. Worth reading end-to-end: `add_new_*`, `unload`, and its helpers.
5. `symbols.rs` — pretty much every former Symbol method except for the ones above
6. One build phase end-to-end (e.g. `python_arch_builder.rs` + its hooks) to see
   call-site ergonomics in practice.
7. LSP features — mostly mechanical conversions.
8. Tests — mostly mechanical conversions.


server/src/weak_collections.rs
```diff
@@ -0,0 +1,83 @@
+use std::{cell::RefCell, collections::{HashMap, HashSet, hash_map}, hash::Hash, vec::IntoIter};
+
+use crate::core::symbols::symbol_keys::KeyValidator;
+
+#[derive(Debug, Clone)]
+pub struct WeakSet<K: Eq + Hash + Copy> {
```
fda-odoo, line 6: maybe call it WeakTableKeySet?

server/src/weak_collections.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub fn retain(&mut self, f: impl Fn(&K) -> bool) {
+        self.set.get_mut().retain(|k| f(k));
+    }
+
+    pub fn drain_valid(&mut self, table: &impl KeyValidator<K>) -> IntoIter<K> {
+        let set = self.set.get_mut();
+        let valid: Vec<K> = set.iter().copied().filter(|k| table.is_key_valid(*k)).collect();
+        set.clear();
+        valid.into_iter()
```
fda-odoo, line 58: this function seems unperformant?
You are copying content in a new vec you collect, then clear the set.
If you want to return owned value, maybe we could drain first then iter.filter on it, to avoid copy?
```rust
pub fn drain_valid(&mut self, table: &impl KeyValidator<K>) -> impl Iterator<Item = K> + '_ {
    self.set.get_mut().drain().filter(|k| table.is_key_valid(*k))
}
```

server/src/weak_collections.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub fn remove(&mut self, key: &K) -> Option<V> {
+        self.map.remove(key)
+    }
+
+    /// iterate over values of valid keys
+    pub fn iter_valid_values(&mut self, table: &impl KeyValidator<K>) -> IntoIter<&V> {
+        self.map.retain(|k, _| table.is_key_valid(*k));
+        let snapshot: Vec<_> = self.map.values().collect();
+        snapshot.into_iter()
```
fda-odoo: what's the call of the snapshot here?
It seems useless copies too
What about
```rust
pub fn iter_valid_values(&mut self, table: &impl KeyValidator<K>) -> impl Iterator<Item = &V> {
    self.map.retain(|k, _| table.is_key_valid(*k));
    self.map.values()
}
```

server/src/core/python_arch_builder.rs
```diff
@@ -958,7 +956,7 @@ impl PythonArchBuilder {
                     body_version_ok = true;
                 }
                 self.visit_node(session, &if_stmt.body)?;
-                vec![ SectionIndex::INDEX(scope.borrow().as_symbol_mgr().get_last_index())]
+                vec![ SectionIndex::INDEX(st!().as_symbol_mgr(scope).get_last_index())]
```
fda-odoo: prefix_section could be used instead of retriveing the symbol again

server/src/core/odoo.rs
```diff
@@ -363,22 +369,71 @@ impl SyncOdoo {
...
             SyncOdoo::build_modules(session);
         }
-    }
+
+        // Self::log_counts(session);
+        // Self::print_memory();
+        // Self::log_capacities(session);
+    }
+
+    // fn log_counts(session: &mut SessionInfo) {
```
fda-odoo: we can keep this kind of metrics. You can maybe organize them in a 'metric' file, maybe guarded by a rust feature or constant in constant.rs?

server/src/core/odoo.rs
```diff
@@ -526,41 +578,37 @@ impl SyncOdoo {
...
-                            Err(_) => {}
+        let Some(addons_symbol) = session.sync_odoo.get_symbol(
+            session.sync_odoo.config.odoo_path.as_ref().unwrap(), &tree(vec!["odoo", "addons"], vec![]), u32::MAX
+        ).first().copied() else {
+            let message = S!("OdooLS: Unable to find 'odoo/addons'. Check the addons_paths in your config or your file structure. Skipping addons loading...");
+            warn!("{}", message);
+            session.show_message(MessageType::WARNING, message);
+            return;
+        };
+        // @arena: not in the original code. Consider wrapping get_symbol in a get_addon_namespace function
```
fda-odoo: maybe remove this comment?

server/src/core/odoo.rs
```diff
@@ -629,96 +677,90 @@ impl SyncOdoo {
...
+        let mut selected_count: u32 = 999999999;
+        let mut current_count: u32;
+        for sym in set.iter_valid(&self.symbol_table) {
+            current_count = 0;
+            let file = self.symbol_table.get_file(sym).unwrap();
+            let all_dep = self.symbol_table.get_all_dependencies(file, step);
+            if let Some(all_dep) = all_dep {
+                for (index, dep_set) in all_dep.iter().enumerate() {
+                    if let Some(dep_set) = dep_set {
+                         let index_set =  match index {
```
fda-odoo: there is spaces and wrong indentation here

server/src/core/odoo.rs
```diff
@@ -629,96 +677,90 @@ impl SyncOdoo {
...
-                        }
-                    }
-                }
-                if current_count < selected_count {
-                    selected_sym = Some(sym.clone());
-                    selected_count = current_count;
-                    if current_count == 0 {
-                        break;
+                        current_count +=
+                            dep_set.iter_valid(&self.symbol_table)
```
fda-odoo: put dep_set.iter.... on previous line

server/src/core/symbols/symbol_keys.rs
```diff
@@ -0,0 +1,225 @@
...
+            _ => panic!("Not a XmlFileKey"),
+        }
+    }
+    
+    pub fn unwrap_csv_file_key(&self) -> CsvFileKey {
+        match self {
+            SymbolKey::CsvFile(k) => *k,
+            _ => panic!("Not a CsvFileKey"),
+        }
+    }
```
mmahrouss, line 120: Perhaps we could avoid some code duplication here, we already use a lot of macros, do you think this also warrants one ?

fda-odoo: I'm not really fan of adding macros everywhere personally

server/src/weak_collections.rs
cammarosano: you're right! no need to copy here, and the current callers does not need to own a copy either.

server/src/core/symbols/symbol_keys.rs
cammarosano: I would look something like this:
```rust
// ...existing code...

macro_rules! impl_unwrap_key {
    ($($method:ident, $variant:ident, $key_type:ty),* $(,)?) => {
        $(
            pub fn $method(&self) -> $key_type {
                match self {
                    SymbolKey::$variant(k) => *k,
                    _ => panic!("Not a {}", stringify!($key_type)),
                }
            }
        )*
    };
}

impl SymbolKey {
    // ...existing code (typ())...

    impl_unwrap_key! {
        unwrap_function_key,      Function,      FunctionKey,
        unwrap_variable_key,      Variable,      VariableKey,
        unwrap_class_key,         Class,         ClassKey,
        unwrap_file_key,          File,          FileKey,
        unwrap_python_package_key,PythonPackage, PythonPackageKey,
        unwrap_module_key,        Module,        ModuleKey,
        unwrap_namespace_key,     Namespace,     NamespaceKey,
        unwrap_root_key,          Root,          RootKey,
        unwrap_xml_file_key,      XmlFile,       XmlFileKey,
        unwrap_csv_file_key,      CsvFile,       CsvFileKey,
    }
}

// ...existing code...
```
Perso think having a macro defining a function (including its name) is a little too much.
I fell it's a bit more than just implementing a trait's method.

server/src/core/model.rs
```diff
@@ -72,85 +71,94 @@ impl ModelData {
...
-            if !sym.borrow().as_class_sym()._model.as_ref().unwrap().inherit.contains(&sym.borrow().as_class_sym()._model.as_ref().unwrap().name) {
-                if from_module.is_none() || sym.as_ref().borrow().find_module().is_none() {
-                    res.push(sym);
+    pub fn get_main_symbols(&self, session: &SessionInfo, from_module: Option<ModuleKey>) -> Vec<ClassKey> {
+        let st = &session.sync_odoo.symbol_table;
+        let mut res = vec![];
+        for key in self.symbols.iter_valid(st) {
+            let model = st[key]._model.as_ref().unwrap();
+            if !model.inherit.contains(&model.name) {
+                let module = st.find_module(key);
```
fda-odoo: find_module is browsing the symbol tree, it could be nice to keep it behind from_module.is_none() for performances

server/Cargo.toml
```diff
@@ -54,5 +55,5 @@ gungraun = "0.17.2"
 assert_fs = "1.1"
 
 [features]
-default = []
+default = ["debug_yarn"]
```
fda-odoo: don't forget to remove that before merge (reminder)

server/src/core/symbols/storage/xml_file_symbol.rs
```diff
@@ -63,86 +56,37 @@ impl XmlFileSymbol {
...
-        &mut self.dependents
-    }
-
-    pub fn is_in_workspace(&self) -> bool {
-        self.in_workspace
+    fn set_build_status(&mut self, step: BuildSteps, status: BuildStatus) {
+        match step {
+            BuildSteps::SYNTAX => panic!(),
+            BuildSteps::ARCH => self.arch_status = status,
+            BuildSteps::ARCH_EVAL => {},
```
mmahrouss, line 81: Any reason why here you do nothing, but in Csv symbol you panic?

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -0,0 +1,444 @@
...
+        let is_external = self.is_external(parent);
+        let compiled_symbol = CompiledSymbol::new(name, path, parent, is_external);
+        let compiled_key = self.compiled.insert(compiled_symbol);
+        match parent {
+            SymbolKey::Compiled(c) => {
+                let replaced_key = self.compiled[c].module_symbols.insert(oyarn!("{}", name), compiled_key.into());
+                if let Some(replaced) = replaced_key {
+                    self.remove(replaced);
+                }
+            },
```
mmahrouss: Do you think it will be more uniform and readible if we move this section to `add_to_parent_module_symbols` and check that child is a compiled key? So we add this section
```rust 
            SymbolKey::Compiled(c) if matches!(child.typ(), SymType::COMPILED) => {
                self.compiled[c].module_symbols.insert(oyarn!("{}", name), child)
            },
```

Then in `add_new_compiled` we call `add_to_parent_module_symbols` normally as in the other add functions? 

If you disagree, I think it will be worth adding a comment saying that a compiled parent can only have compiled child symbols?

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+    pub fn add_new_function(&mut self, parent: SymbolKey, name: &str, range: &TextRange, body_start: &TextSize) -> FunctionKey {
+        let is_external = self.is_external(parent);
+        let function_symbol = FunctionSymbol::new(name, parent, range.clone(), body_start.clone(), is_external);
+        let function_key = self.functions.insert(function_symbol);
+        self.add_to_parent_symbols(parent, function_key.into(), name, range.start().to_u32());
+        function_key
+    }
+
+    pub fn add_new_class(&mut self, parent: SymbolKey, name: &String, range: &TextRange, body_start: &TextSize) -> ClassKey {
```
mmahrouss: `&String` -> `&str`

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -0,0 +1,444 @@
...
+        let variable_key = self.variables.insert(variable_symbol);
+        let section = self.get_section_for_key(owner, range.start().to_u32());
+    
+        self.ext_symbols.add(target, owner, name, section, variable_key);
+        variable_key
+    }
+
+    // ====== Helpers for symbol creation ======
+
+    // If replacing an exisiting entry, the old one gets removed from the symbol table to prevent leaks,
```
mmahrouss: `exisiting` -> `existing`

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -0,0 +1,444 @@
...
+            SymbolKey::PythonPackage(p) => self[p].get_section_for(position).index,
+            SymbolKey::Class(c) => self[c].get_section_for(position).index,
+            SymbolKey::Function(f) => self[f].get_section_for(position).index,
+            _ => panic!("Impossible to add a declaration of external symbol to a {}", owner.typ()),
+        }
+    }
+    
+    // ====== Symbol removal ======
+ 
+    /// Remove `symbol` and its descendents, and clean up.
```
mmahrouss: `descendents` -> `descendants`

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// WARNING: After this call, `symbol` and its descendents are no longer valid keys.
+    pub fn unload(session: &mut SessionInfo, symbol: SymbolKey) {
+        fn unload_recursively(session: &mut SessionInfo, symbol: SymbolKey) {
+            for child in session.st().children(symbol) {
+                unload_recursively(session, child);
+            }
+            SymbolTable::on_unload(session, symbol);
+        }
+        unload_recursively(session, symbol);
```
mmahrouss, line 277: love the recursion <3

server/src/features/xml_ast_utils.rs
```diff
@@ -121,21 +131,21 @@ impl XmlAstUtils {
...
+        ctxt.remove("field_name");
     }
 
-    fn visit_text(session: &mut SessionInfo, node: &Node, offset: usize, from_module: Option<Rc<RefCell<Symbol>>>, ctxt: &mut HashMap<String, ContextValue>, results: &mut (Vec<XmlAstResult>, Option<Range<usize>>), on_dep_only: bool) {
+    fn visit_text(session: &mut SessionInfo, node: &Node, offset: usize, from_module: Option<ModuleKey>, ctxt: &mut HashMap<String, ContextValue>, results: &mut (Vec<XmlAstResult>, Option<Range<usize>>), on_dep_only: bool) {
         if node.range().start <= offset && node.range().end >= offset {
-            let model = ctxt.get(&S!("record_model")).cloned().unwrap_or(ContextValue::STRING(S!(""))).as_string();
-            let field = ctxt.get(&S!("field_name")).cloned().unwrap_or(ContextValue::STRING(S!(""))).as_string();
+            let model = ctxt.get("record_model").map(|v| v.as_string()).unwrap_or_default();
+            let field = ctxt.get("field_name").map(|v| v.as_string()).unwrap_or_default();
```
mmahrouss: Maybee, since we are just creating a closure to call a function, we just pass the function, and I think it is also as readable 
```suggestion
            let model = ctxt.get("record_model").map(ContextValue::as_string).unwrap_or_default();
            let field = ctxt.get("field_name").map(ContextValue::as_string).unwrap_or_default();
```

server/src/core/evaluation.rs
```diff
@@ -148,23 +134,16 @@ impl ContextValue {
         }
     }
 
-    pub fn as_string(&self) -> String {
+    pub fn as_string(&self) -> &str {
```
mmahrouss: Since we changed the return type, how about also changing the method name, `as_str`, which is what is commenly used for methods from &self -> &str
```suggestion
    pub fn as_str(&self) -> &str {
```

server/src/core/symbols/storage/xml_file_symbol.rs
cammarosano: That's a great question! ^^
It simply comes from the original `Symbol::set_build_status`:
```rust
     ...
            Symbol::XmlFileSymbol(x) => {
                match step {
                    BuildSteps::SYNTAX => panic!(),
                    BuildSteps::ARCH => x.arch_status = status,
                    BuildSteps::ARCH_EVAL => {},
                    BuildSteps::VALIDATION => x.validation_status = status,
                }
            },
            Symbol::CsvFileSymbol(c) => {
                match step {
                    BuildSteps::SYNTAX => panic!(),
                    BuildSteps::ARCH => c.arch_status = status,
                    BuildSteps::ARCH_EVAL => panic!(),
                    BuildSteps::VALIDATION => c.validation_status = status,
                }
            },
      ...
```
So, I did not really think about it when converting, just copy-pasted. I avoided behavior changes in general...

server/src/core/symbols/storage/lifecycle.rs
cammarosano: I don't speek Englich apperantly!!!! ^^'
(thanks, do keep going)

server/src/core/symbols/storage/lifecycle.rs
cammarosano: Most definitely yes!! (as in: agreed!)
Pushed [REF add_new_csv_file and add_to_module_data_symbols](https://github.com/odoo/odoo-ls/pull/581/commits/7ba38a6f7e1b03e7e933039f062b44d559e773f7)

server/src/core/symbols/storage/lifecycle.rs
cammarosano: too much punk rock in my childhood: https://en.wikipedia.org/wiki/Descendents

server/src/core/evaluation.rs
cammarosano: Yeah, definitely!

server/src/features/xml_ast_utils.rs
cammarosano: YES!

server/Cargo.toml
cammarosano: oops :grimacing:

server/src/core/model.rs
cammarosano: true!
pushed [Optimize Model::get_main_symbols](https://github.com/odoo/odoo-ls/pull/581/commits/13d9d6d83d1aece6c1943dbc34c2361b44325f58)

server/src/features/completion.rs
```diff
@@ -44,7 +52,7 @@ pub struct CompletionFeature;
 impl CompletionFeature {
 
     pub fn autocomplete(session: &mut SessionInfo,
-        file_symbol: &Rc<RefCell<Symbol>>,
+        file_symbol: SymbolKey,
```
mmahrouss: Do you think we can convert it here to FileKey, and in similar situations in features? or is it not worth it ? or is it already planned ?

server/src/features/references_xml.rs
```diff
@@ -78,61 +89,56 @@ impl XmlAstReferenceVisitor {
...
-                let ReferenceTarget::Symbol(target) = target else {continue;};
-                if target.borrow().typ() != SymType::VARIABLE {continue;}
-                if !target.borrow().is_field(session) {continue;}
-                if target.borrow().name() != attr.value() {continue;}
+                let &ReferenceTarget::Symbol(SymbolKey::Variable(target)) = target else {continue;};
+                if !SymbolTable::is_field(session, target.into()) {continue;}
+                if session.st()[target].name != attr.value() {continue;}
                 //field name matches, but we still have to check model is the same
-                let model_name = ctxt.get(&S!("record_model")).cloned().unwrap_or(ContextValue::STRING(S!(""))).as_string();
+                let model_name = ctxt.get("record_model").map(|v| v.as_str()).unwrap_or_default();
```
mmahrouss: Suggestions: 
```suggestion
                let model_name = ctxt.get("record_model").map(ContextValue::as_str).unwrap_or_default();
```
It is also in some places.

server/src/features/xml_ast_utils.rs
```diff
@@ -145,72 +155,69 @@ impl XmlAstUtils {
...
+                    if let Some(file) = file_weak.upgrade(st) {
+                        if let Some(module) = st.find_module(file) {
+                            return ModuleSymbol::is_in_deps(st, st.find_module(file_symbol).unwrap(), &st[module].name);
                         }
                     }
-                        return false;
                 }
-            ).collect::<Vec<_>>();
+                false
+            }).collect::<Vec<_>>();
```
mmahrouss: Style suggestion:) 
```suggestion
            xml_ids = xml_ids.into_iter().filter(|x|
                if let Some(file_weak) = x.get_file_symbol(st)
                && let Some(file) = file_weak.upgrade(st)
                && let Some(module) = st.find_module(file) {
                    ModuleSymbol::is_in_deps(st, st.find_module(file_symbol).unwrap(), &st[module].name)
                } else {
                    false
                }
            ).collect::<Vec<_>>();
```

server/src/core/odoo.rs
cammarosano: Done! See [metrics module](https://github.com/odoo/odoo-ls/pull/581/commits/ca0dcb7bb1264d7ade1bdf19fc71603a4ea315e2) commit

server/src/features/completion.rs
cammarosano: yes! But, if I got this right, it might also be a Module or a Package key.

See [Introduce SourceFileKey](https://github.com/odoo/odoo-ls/pull/581/commits/db44b5e6892f35ddb2a66dd9680870a610be7e65)

server/src/features/references_xml.rs
cammarosano: Nice! [Replace closures in .map by ContextValue::as_str](https://github.com/odoo/odoo-ls/pull/581/commits/0a6d5d8a52b6ec7a05561d05bfbdbccf7eded704)

server/src/features/xml_ast_utils.rs
cammarosano: Stylish.
[Re-style add_xml_id_result by MDMS](https://github.com/odoo/odoo-ls/pull/581/commits/28235a593809e8d4d4b23fb843933531f970e9a9)

server/src/features/completion.rs
mmahrouss: Yes you are right, SourceFileKey handles it beautifully

mmahrouss (approved):

server/src/core/entry_point.rs
```diff
@@ -214,7 +214,7 @@ impl EntryPointMgr {
                 SymbolKey::Namespace(n) => {
                     if file_path.ends_with("__manifest__.py") {
                         warn!("new custom entry point for manifest without related init.py is not supported outside of main entry point. skipping...");
-                        session.sync_odoo.entry_point_mgr.borrow_mut().remove_entries_with_path(tree_path);
+                        session.sync_odoo.entry_point_mgr.borrow_mut().remove_entries_with_path(&mut session.sync_odoo.symbol_table, tree_path);
```
fda-odoo, line 210: session.st_mut() ?

fda-odoo:
@cammarosano I pushed a commit with my comments regarding arena-review.

cammarosano:
Benchmark of [[REF] store symbols in Symbol Table](https://github.com/odoo/odoo-ls/pull/581/commits/5772a2c80af2a9fba2b2eb5fade86e8b82a04e8c) against alpha-1.5.0. CLI mode on Community only (master/19.3):
```
commit                  user (s)             sys (s)       total CPU (s)         peak RSS (MB)
──────────  ──────────────────  ──────────────────  ──────────────────  ────────────────────
3ecb974e         19.410 ± 0.063      2.594 ± 0.046     22.004 ± 0.030         2107.5 ± 0.2
5772a2c8         16.936 ± 0.082      2.324 ± 0.015     19.260 ± 0.083         1660.1 ± 0.1

Δ (after − before):
  user (s):              -2.474  (-12.75%)   z=23.8
  sys (s):               -0.270  (-10.41%)   z=5.6
  total CPU (s):         -2.744  (-12.47%)   z=31.2
  peak RSS (MB):         -447.4  (-21.23%)   z=2061.8

```

Raw data:
```
Benchmark: 3ecb974e  →  5772a2c8
Runs: 5 (plus 1 warm-up each)
Current branch: HEAD
Per-run logs:   /tmp/odools_bench_454766/

>>> Checking out 3ecb974e...
>>> Building release...

==> Warming up (3ecb974e)...
    warm-up done (user=19.49s sys=2.62s rss=2158064kB)
    run 1: user=19.32s sys=2.66s wall=21.98s rss=2158212kB
    run 2: user=19.44s sys=2.54s wall=21.99s rss=2158080kB
    run 3: user=19.40s sys=2.59s wall=21.98s rss=2158212kB
    run 4: user=19.38s sys=2.63s wall=22.02s rss=2157684kB
    run 5: user=19.51s sys=2.55s wall=22.06s rss=2157968kB

>>> Checking out 5772a2c8...
>>> Building release...

==> Warming up (5772a2c8)...
    warm-up done (user=16.93s sys=2.28s rss=1699820kB)
    run 1: user=16.95s sys=2.31s wall=19.26s rss=1699916kB
    run 2: user=16.89s sys=2.31s wall=19.19s rss=1699996kB
    run 3: user=16.82s sys=2.33s wall=19.15s rss=1699852kB
    run 4: user=17.07s sys=2.32s wall=19.38s rss=1699740kB
    run 5: user=16.95s sys=2.35s wall=19.29s rss=1700032kB

 Raw data points                                                

  Before (3ecb974e)
    run        user(s)     sys(s)    wall(s)        rss(MB)
    1            19.32       2.66      21.98         2107.6
    2            19.44       2.54      21.99         2107.5
    3            19.40       2.59      21.98         2107.6
    4            19.38       2.63      22.02         2107.1
    5            19.51       2.55      22.06         2107.4

  After  (5772a2c8)
    run        user(s)     sys(s)    wall(s)        rss(MB)
    1            16.95       2.31      19.26         1660.1
    2            16.89       2.31      19.19         1660.2
    3            16.82       2.33      19.15         1660.0
    4            17.07       2.32      19.38         1659.9
    5            16.95       2.35      19.29         1660.2

```

PR #583, opened 2026-04-15
[FIX] Use safe get xml symbol file in go to utils

mmahrouss:
For crash 2114


fda-odoo (approved):

fda-odoo:
@fda-odoo include this in OdooData refactoring

PR #585, opened 2026-04-15
[IMP] Avoid sending error during shutdown

mmahrouss:
The aim of this PR, is to not panic because another thread has been shutdown, due to odooLS being disabled or during a normal shutdown sequence.

### DIFF
--- a/server/src/threads.rs
+++ b/server/src/threads.rs
@@ -34,12 +34,16 @@ impl <'a> SessionInfo<'a> {
         }
     }
     pub fn log_message(&self, msg_type: MessageType, msg: String) {
-        self.sender.send(
+        if let Err(e) = self.sender.send(
             Message::Notification(lsp_server::Notification{
                 method: LogMessage::METHOD.to_string(),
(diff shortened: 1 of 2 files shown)


server/src/threads.rs
```diff
@@ -48,30 +52,40 @@ impl <'a> SessionInfo<'a> {
...
-        self.sender.send(
+        if let Err(_) = self.sender.send(
             Message::Notification(lsp_server::Notification{
                 method: ShowMessage::METHOD.to_string(),
                 params: serde_json::to_value(&ShowMessageParams{typ: msg_type, message: msg}).unwrap()
             })
-        ).unwrap();
+        ) {
+            if !self.sync_odoo.terminate_rebuild.load(std::sync::atomic::Ordering::SeqCst) {
+                panic!("Failed to send show_message - server not shutting down");
```
fda-odoo: Maybe change the message? I thought at first that the error is that the server is not shutting down.
"Failed to send show_message, but server is not shutting down" ?

mmahrouss: Yes, that is clearer indeed

fda-odoo (approved):

PR #587, opened 2026-04-21
[REF] Refactor config into a module

mmahrouss:
Check the readme.md in the conifg module for more information


server/src/core/config/loading.rs
```diff
@@ -0,0 +1,667 @@
...
+        HashMap::new(),
+        &config_dir,
+    ) {
+        Ok(filled_path) => {
+            let var_pb = PathBuf::from(&filled_path);
+            if var_pb.is_file()
+            && let Some(file_name) = var_pb.file_name()
+            && file_name.to_string_lossy() == "__manifest__.py"
+            {
+                let contents =  match fs::read_to_string(&var_pb) {
```
fda-odoo: you should maybe use fileInfo cache so we keep using the same tools and caching for parsing files?

server/src/core/config/loading.rs
```diff
@@ -0,0 +1,667 @@
...
+    apply_extends(&mut config_map)?;
+    // Process vars (skip abstract configs)
+    config_map
+        .values_mut()
+        .filter(|e| !e.abstract_)
+        .for_each(|entry| {
+            entry.version = entry
+                .version
+                .clone()
+                .map(|v| process_version(v, unique_ws_folders, current_ws));
```
fda-odoo: I'm lost here. Why are you searching for `__manifest__.py` version? What's the purpose of the version?

server/src/core/config/merge.rs
```diff
@@ -0,0 +1,385 @@
...
+pub(super) fn apply_merge(child: &ConfigEntryRaw, parent: &ConfigEntryRaw) -> ConfigEntryRaw {
+    let mut result = ConfigEntryRaw::default();
+    result.name = child.name.clone();
+    result.extends = child.extends.clone().or(parent.extends.clone());
+
+    // Scalar fields: child value wins, fallback to parent
+    result.merge_scalars_from(child, parent);
+
+    // Special scalar fields
+    result.odoo_path = child.odoo_path.clone().or(parent.odoo_path.clone());
```
fda-odoo: maybe we shouldn't keep all variables in the struct, but use a hashmap where we would store all values, the napply same generic functions on it? It would make addition way more easy (I think we'll have to add dozen of new configs values in the future...)
What do you think?

fda-odoo (changes requested):

server/src/core/config/loading.rs
mmahrouss: We have a feature where you can set the `$version` variable to point to a manifest file. and in doing so we read the manifest file and extract the version from there

server/src/core/config/merge.rs
mmahrouss: But this is the point of having normal scalar fields, you will not need to add any new code, just adding them in `schema.rs` under `scalars { ... }` 

However there are special parameters, that need custom behavior `odoo_path, python_path, addons_paths, additional_stubs, additional_languages, diagnostic_settings, diagnostic_filters, version, base` 

We can expand the macro usage to also cover the mentioned fields in places where they are generic, so it will have less boilerplate. 

But the result anyway, is that adding a simple config param will just be as easy as just defining it in the `schema.rs`

server/src/core/config/loading.rs
mmahrouss: The issue is that this runs before any files are parsed, so it is simpler to read one file. Also, a user can just point to any manifest file in principle. But it is always before loading, and we do not know addons paths or odoo paths yet.

mmahrouss: I think the overlap is very small, to justify linking it to file info, the point is to have a very small parser, to read the file and extract the version.

We could mimic opening the file, have the session passed through, but I think it is not complex enough to justify that, what do you think?

.gitignore
```diff
@@ -9,3 +9,4 @@ server/cross/
 *.log
 *.vsix
 *.lock
+.claude
```
cammarosano: considering `.claude` dir does not exist in our repo, you should maybe add this to your `.git/info/exclude` instead. (I'ts like a personal .gitignore)

server/src/core/config/spec.rs
```diff
@@ -0,0 +1,239 @@
...
+        self
+    }
+    pub(crate) fn symlink_policy(mut self, policy: SymlinkPolicy) -> Self {
+        self.symlink_policy = policy;
+        self
+    }
+    pub(crate) fn stage(mut self, stage: Stage, run: HookFn) -> Self {
+        self.stages.push(StageHook { stage, run });
+        self
+    }
```
cammarosano, line 149: this is awesome!

server/src/core/config/spec.rs
```diff
@@ -0,0 +1,239 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// `build_registry` panics at `LazyLock` init on a duplicate key or name.
+    /// Touch the registry here so that programmer error surfaces in CI rather
+    /// than only at first runtime use. Also pins that every `ConfigKey` has a
+    /// spec (so the accessors below cannot panic).
+    #[test]
+    fn registry_builds_and_covers_every_key() {
```
cammarosano, line 229: golden

server/src/core/config/value.rs
```diff
@@ -0,0 +1,472 @@
+//! The config **data model**: the runtime value types and the per-profile state
+//! the pipeline operates on.
+//!
+//! Layered below `spec.rs` and `build.rs` (it depends on neither): public value
+//! enums, the `Sourced<T>` provenance wrapper and its merge helpers, the
+//! internal `Scalar`/`ConfigValue` representation, and the `Profile` working
+//! structure. The field-declaration machinery (`FieldSpec`, the registry) lives
+//! in `spec.rs`.
+// TODO: continue the review here
```
cammarosano: ?

server/src/core/config/stages.rs
```diff
@@ -0,0 +1,457 @@
...
+                    ctx.unique_ws_folders,
+                    ctx.current_ws,
+                    &raw,
+                    vars.clone(),
+                )
+                .and_then(|p| glob::Pattern::new(&p).map_err(|e| e.to_string()));
+                match result {
+                    Ok(p) => Some(p),
+                    Err(err) => {
+                        info_add.push_str(&format!("Failed to process pattern '{raw}': {err}"));
```
cammarosano: no separator?

cammarosano (commented):
Very nice! Super well-structured and well documented.
I did not do an in-depth review. But the new API for registering new config keys/fields is amazing!

server/src/core/config/value.rs
mmahrouss: oops, I will remove that:) it was a note for me while self-reviewing

server/src/core/config/spec.rs
mmahrouss: builder pattern!:)

server/src/core/config/stages.rs
mmahrouss: I am not sure I understand what separator you think is missing? like a new line char? or like code formatting?

fda-odoo (approved):
Good for me ! Well done with the result !
It's really well improved and will easily allow addition of new config entries !

cammarosano (approved):

PR #588, opened 2026-04-22
Alpha xml new sym fda

fda-odoo:


fda-odoo:
@mmahrouss tests are passing after the rebase, but I didn't test it deeply though. You can still start the review however

server/src/core/symbols/storage/xml/xml_asset_symbol.rs
mmahrouss, line 1: I see `server/src/core/symbols/storage/xml/xml_file_symbol.rs` is formatted, but not `server/src/core/symbols/storage/xml/xml_asset_symbol.rs` and `server/src/core/symbols/storage/xml/xml_delete_symbol.rs` 

I would run rustfmt in all files (or none) especially for new files.

server/src/core/symbols/storage/module_symbol.rs
```diff
@@ -102,6 +102,10 @@ impl ModuleSymbol {
         module.load(session, dir_path)
     }
 
+    pub fn insert_xml_id(symbol_table: &mut SymbolTable, target: ModuleKey, xml_id: OYarn, xml_data: XmlId) {
+        symbol_table[target].xml_ids.entry(xml_id).or_insert(WeakSet::new()).insert(xml_data);
+    }
+
```
mmahrouss: I think should live in module_load, as it is not concerning parent/child relationship. Do you agree? 

See commit message here https://github.com/odoo/odoo-ls/commit/39a93c499bdb4ace12789a532cf2815e4dce454a

server/src/core/csv_arch_builder.rs
```diff
@@ -71,9 +69,8 @@ impl CsvArchBuilder {
                                 csv_module = module;
                             }
                         }
-                        session.st_mut()[csv_module].xml_id_locations.entry(Sy!(id_split.last().unwrap().to_string())).or_default().insert(csv_symbol.into());
-                        data_hooks::on_record_creation(session, csv_symbol.into(), &record);
-                        session.st_mut()[csv_symbol].xml_ids.entry(Sy!(id_split.last().unwrap().to_string())).or_insert(vec![]).push(OdooData::RECORD(record));
+                        data_hooks::on_record_creation(session, csv_symbol.into(), &record_key);
+                        session.st_mut()[csv_module].xml_ids.entry(Sy!(id_split.last().unwrap().to_string())).or_insert(WeakSet::new()).insert(XmlId::XmlRecord(record_key));
```
mmahrouss:
```suggestion
                        ModuleSymbol::insert_xml_id(session.st_mut(), csv_module, Sy!(id_split.last().unwrap().to_string()), XmlId::XmlRecord(record_key));
```

Use the method `insert_xml_id` you already created ;)

server/src/core/xml_arch_builder.rs
```diff
@@ -83,21 +80,20 @@ impl XmlArchBuilder {
...
             }
-            xml_data.set_file_symbol(self.xml_symbol);
-            session.st_mut()[xml_module].xml_id_locations.entry(Sy!(id.clone())).or_default().insert(self.xml_symbol.into());
-            if let OdooData::RECORD(ref record) = xml_data {
-                data_hooks::on_record_creation(session, self.xml_symbol.into(), record);
+            if let XmlDataKey::RECORD(record) = xml_data {
+                data_hooks::on_record_creation(session, self.xml_symbol.into(), &record);
             }
-            session.st_mut()[self.xml_symbol].xml_ids.entry(Sy!(id)).or_insert(vec![]).push(xml_data);
+            session.st_mut()[xml_module].xml_ids.entry(Sy!(id)).or_insert(WeakSet::new()).insert(XmlId::from(xml_data));
```
mmahrouss: Same comment as https://github.com/odoo/odoo-ls/pull/588/changes#r3160985839

server/src/core/symbols/storage/package_symbol.rs
```diff
@@ -1,6 +1,6 @@
 use weak_table::PtrWeakHashSet;
 
-use crate::{constants::{BuildStatus, BuildSteps, OYarn}, core::{file_mgr::NoqaInfo, model::Model, symbols::{storage::dependency_mgr::{DependenciesTable, DependentsTable}, symbol_keys::SymbolKey}, xml_data::OdooData}, oyarn, utils::{NoHashBuilder, PathSanitizer}};
+use crate::{constants::{BuildStatus, BuildSteps, OYarn}, core::{file_mgr::NoqaInfo, model::Model, symbols::{storage::dependency_mgr::{DependenciesTable, DependentsTable}, symbol_keys::{SourceFileKey, SymbolKey}}}, oyarn, utils::{NoHashBuilder, PathSanitizer}, weak_collections::WeakSet};
```
mmahrouss: Reminder ;) Check warnings from cargo check, unused imports, and also in other files.

server/src/core/symbols/module_load.rs
```diff
@@ -516,7 +522,7 @@ impl ModuleSymbol {
         data_hooks::on_file_unload(session, data_file);
     }
 
-    fn load_arch(module_key: ModuleKey, session: &mut SessionInfo) -> Vec<Diagnostic> {
+    fn _load_arch(module_key: ModuleKey, session: &mut SessionInfo) -> Vec<Diagnostic> {
```
mmahrouss:
```suggestion
    fn load_arch(module_key: ModuleKey, session: &mut SessionInfo) -> Vec<Diagnostic> {
```

I would keep the _ out, I guess this was a conflict resolution issue. But we already have `pub` the `_` is unnecessary. same for `_load_depends` and `_load_manifest`
What do you think ?

server/src/core/symbols/module_load.rs
```diff
@@ -23,7 +23,11 @@ use tracing::{error, info};
 impl ModuleSymbol {
 
     pub fn load(mut self, session: &mut SessionInfo, dir_path: &PathBuf) -> Option<Self> {
-        info!("building new module: {:?}", self.path);
+        info!("building new module: {:?}", dir_path.sanitize());
+        if dir_path.components().last().unwrap().as_os_str().to_str().unwrap() == "base" {
+            self.depends.clear();
+        }
+        self.dir_name = oyarn!("{}", dir_path.with_extension("").components().last().unwrap().as_os_str().to_str().unwrap());
```
mmahrouss: Isn't that a duplicate of the beginning of `ModuleSymbol::new`?

I think it is a conflict issue, see https://github.com/odoo/odoo-ls/commit/39a93c499bdb4ace12789a532cf2815e4dce454a

server/src/core/symbols/module_load.rs
```diff
@@ -36,23 +40,23 @@ impl ModuleSymbol {
...
-        let diags = self.load_manifest(session, &manifest_file_info);
+        let diags = self._load_manifest(session, &manifest_file_info);
         if session.sync_odoo.modules.contains_key(&self.dir_name) {
             //TODO: handle multiple modules with the same name
         }
         manifest_file_info.replace_diagnostics(crate::constants::BuildSteps::SYNTAX, diags);
         manifest_file_info.publish_diagnostics(session);
         drop(manifest_file_info);
-        info!("End building new module: {:?}", self.path);
+        info!("End building new module: {:?}", dir_path.sanitize());
```
mmahrouss:
```suggestion
        info!("End building new module: {:?}", self.path);
```
More conflict issues in that file.

server/src/features/csv_ast_utils.rs
```diff
@@ -154,7 +153,7 @@ impl CsvAstUtils {
         record: &StringRecord,
         results: &mut Vec<GotoSource>,
         csv_symbol: SourceFileKey,
-        model_name: OYarn,
+        _model_name: OYarn,
```
mmahrouss:
```suggestion
        model_name: OYarn,
```

server/src/features/csv_ast_utils.rs
```diff
@@ -172,22 +171,23 @@ impl CsvAstUtils {
...
-                            end,
-                        },
-                    })),
-                    origin_selection_range: None,
-                })
+                for data in session.st()[csv_symbol.unwrap_csv_file_key()].symbols.iter() {
+                    if let Some(record_key) = data.as_xml_record_key() {
+                        let record = &session.st()[record_key];
+                        let Some(record_xml_id) = &record.xml_id else {continue;};
+                        if record.range.start().to_usize() <= start && record.range.end().to_usize() >= end && record_xml_id.as_str() == field_data.as_str() {
```
mmahrouss: Perhaps it will be nice to use `TextRange::contains_range` as it does the exact same check, but with a readable title
and we can create the `field_range` before the loop, just a style suggestion.

server/src/core/symbols/storage/xml/xml_record_symbol.rs
```diff
@@ -0,0 +1,44 @@
...
+use ruff_text_size::TextRange;
+
+use crate::{constants::OYarn, core::symbols::symbol_keys::{SymbolKey, XmlFieldKey}, weak_collections::WeakSet};
+
+#[derive(Debug)]
+pub struct XmlRecordSymbol {
+    pub is_external: bool,
+    pub model: (OYarn, Range<usize>),
+    pub xml_id: Option<OYarn>,
+    pub fields: HashMap<OYarn, WeakSet<XmlFieldKey>>,
```
mmahrouss: Can we have multiple fields with the same name? and why would we want to save them? won't the last one just shadow? or how does it work? 

I think maybe we could add a comment to explain why, if it is needed, or if it is actually not needed then we should change it to WK<XmlFieldKey>

server/src/features/csv_ast_utils.rs
```diff
@@ -172,22 +171,23 @@ impl CsvAstUtils {
...
+                                    results.push(GotoSource {
+                                        source: xml_id.into(),
+                                        origin_selection_range: None,
+                                    });
+                                }
+                                break;
+                            }
+                        }
+                    }
+                }
```
mmahrouss: I feel like this snippet is confusing to read. After spending time with it, I find that it does the following. 
1. Find the one record that includes this field's start and end. (So the loop essentially does logic once, so maybe it should be replaced with a loop)
2. Then on that one record, we check that the id match, and then we look for id fields, which there should be one? 

I suggest we split it into a part that gets the id fields, and a part that adds them into result? with a comment(s).
Here is a suggestion (not AI written), I feel it is more readable? or I could be biased, feel free to dismiss or draft a different way.

I still do not understand why we can have multiple id fields too
```suggestion

                let record_range =
                    TextRange::new(TextSize::from(start as u32), TextSize::from(end as u32));
                // Find the record that corresponds to the current id
                // And extract the field key(s?) for the "id" field
                let maybe_id_field_keys = session.st()[csv_symbol.unwrap_csv_file_key()]
                    .symbols
                    .iter()
                    .filter_map(|s| s.as_xml_record_key())
                    .find_map(|record_key| {
                        let record = &session.st()[record_key];
                        let Some(record_xml_id) = &record.xml_id else {
                            return None;
                        };
                        if record.range.contains_range(record_range)
                            && record_xml_id.as_str() == field_data.as_str()
                        {
                            record.fields.get("id")
                        } else {
                            None
                        }
                    });
                // Add the id field symbols to the results
                let Some(id_field_keys) = maybe_id_field_keys else {
                    continue;
                };
                let sources = id_field_keys
                    .iter_valid(session.st())
                    .map(|xml_id| GotoSource {
                        source: xml_id.into(),
                        origin_selection_range: None,
                    });
                results.extend(sources);
```

server/src/features/goto_utils.rs
```diff
@@ -156,17 +149,15 @@ impl GotoUtils {
...
-                    xml_found = true;
-                    let path = session.st().path(file_symbol).to_string();
-                    sources.push(GotoSource{
-                        source: GotoSourceType::OdooData(xml_id.clone()),
-                        origin_selection_range: eval.range.map(|r| session.sync_odoo.get_file_mgr().borrow().text_range_to_range(session, &path, &r))
-                    });
-                }
+        for xml_id in xml_ids.iter_valid(session.st()) {
+            let file = session.st().get_file(xml_id.into());
+            if let Some(_file) = file {
```
mmahrouss: Suggestion: We can change it to let else continue, so we have less indentation

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -235,6 +289,21 @@ impl SymbolTable {
...
+            },
+            SymbolKey::CsvFile(f) => {
+                let csv_file = &mut self.csv_files[f];
+                let csv_data_key = content.as_xml_data_key().expect("content should be an xml data key");
+                csv_file.symbols.insert(csv_data_key);
+            },
+            SymbolKey::XmlRecord(r) => {
+                let xml_record = &mut self.xml_records[r];
+                let xml_field_key = content.unwrap_xml_field_key();
+                xml_record.fields.entry(oyarn!("{}", name)).or_insert_with(WeakSet::new).insert(xml_field_key);
```
mmahrouss: since the upstream parent, does not have `symbols` with type SymbolKey, it is not under the umbrella of `add_to_parent_symbols` and you can sense it is weird that you are casting the key at the call, then expecting it here. Since it is not a lot of code duplication, we could keep them in their `add_new_xml_...` 

Or create add_to_parent_xml_data_symbols for Xml file and csv file, what do you think?

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -676,7 +677,8 @@ impl XmlArchBuilder {
                     }
                 },
                 "field" => {
-                    self.load_field(session, &child, diagnostics);
+                    //TODO uncomment field loading on assets
+                    //self.load_field(session, &child, diagnostics);
```
mmahrouss: What was the issue with the fields on assets?

server/src/features/workspace_symbols.rs
mmahrouss, line 1: Not related to this PR specifically, but I think we need tests for workspace symbols

server/src/features/references.rs
mmahrouss, line 58: Is this comment still correct?

mmahrouss (approved):
95% style suggestions:) 

My main point that I am confused about is why we have a weakset not a weak key for fields, other than that it is really cool. 
Great PR, we needed it, and will enable us to superpower our CSV/XML features :D ( we will parse XML models soon:))

server/src/features/csv_ast_utils.rs
fda-odoo: I removed the field

server/src/core/symbols/storage/xml/xml_record_symbol.rs
fda-odoo: I changed to only keep the last declared one. It's resolving another comment too

server/src/features/csv_ast_utils.rs
fda-odoo: I only added comments in the code here

server/src/features/goto_utils.rs
fda-odoo: if .is_none() is even better

fda-odoo:
@mmahrouss thank you for the review. I updated the code regarding your comments :)

mmahrouss (changes requested):
At module_load, there are still many changes that are reverting https://github.com/odoo/odoo-ls/commit/39a93c499bdb4ace12789a532cf2815e4dce454a , other than that looks great!, thank you:)

server/src/core/symbols/storage/lifecycle.rs
cammarosano: I agree with @mmahrouss that this is a bit weird. It's also weird at the call site, where you pass an empty string as the `name` param, which is another indicator that this function is not suitable for the job:
```rust
self.add_to_parent_symbols(parent.into(), xml_template_key.into(), "", range.start().to_u32());
```

I think mdms's second suggest would be the cleanest option.
> create add_to_parent_xml_data_symbols for Xml file and csv file

Why? Because it makes the code in this file easier to "audit" for leaks and symbol table integrity.
- Symbol variants that has a `symbols` field with the sections: all grouped in `add_to_parent_symbols`, and new keys get pushed into a vector -> we don't have to worry about overwriting existing keys and have a leak.
- Variants that have the `module_symbols` maps keyed by "path" -> here we have to handle the case in which the value for a given path gets overwritten (we have remove that symbol from the table otherwise we would have a leak)
- the new `symbols` in XmlFile and CsvFile: as this is a set, we would not replace a key by adding it again, so no memory issues here.

So, while your code does not introduce memory leak issues,  it groups 2 data structure "cases" in the same function: one it's a map of map of vectors, and the other is just a map. And keeping stuff separated makes it easier for us to track issues.

server/src/core/symbols/storage/xml/xml_file_symbol.rs
cammarosano, line 68: Please update this to return the keys in the `symbols` HashSet. This is crucial for removing the children symbols from the table when the parent (the xml file symbol) is removed.

server/src/core/symbols/storage/csv_file_symbol.rs
cammarosano: Please update this to return the keys in the symbols HashSet. This is crucial for removing the children symbols from the table when the parent (the csv file symbol) is removed.

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -210,16 +211,16 @@ impl XmlArchBuilder {
...
+            (oyarn!("{}", node.attribute("model").unwrap()), node.attribute_node("model").unwrap().range()),
+            found_id.clone().map(|id| oyarn!("{}", id)),
+            &TextRange::new(TextSize::new(node.range().start as u32), TextSize::new(node.range().end as u32))
+        );
+        let mut fields = vec![];
         for child in node.children().filter(|n| n.is_element()) {
-            if let Some(field) = self.load_field(session, &child, diagnostics) {
-                data.fields.push(field);
+            if let Some(field) = self.load_field(session, &child, record.into(), diagnostics) {
+                fields.push(field);
```
cammarosano: what is the purpose of this `fields` vector?
you push content into it, but you never read it.

server/src/core/symbols/storage/xml/xml_record_symbol.rs
```diff
@@ -0,0 +1,44 @@
...
+use ruff_text_size::TextRange;
+
+use crate::{constants::OYarn, core::symbols::symbol_keys::{SymbolKey, Wk, XmlFieldKey}, weak_collections::WeakSet};
+
+#[derive(Debug)]
+pub struct XmlRecordSymbol {
+    pub is_external: bool,
+    pub model: (OYarn, Range<usize>),
+    pub xml_id: Option<OYarn>,
+    pub fields: HashMap<OYarn, Wk<XmlFieldKey>>,
```
cammarosano: why are fields stored here as weak? Are they stored as strong key anywhere else?
If a `XmlRecordSymbol` owns the `XmlFieldSymbol` (parent - child relationship):
- these should be strong keys ( `HashMap<OYarn, XmlFieldKey>`)
- the `children` method should return the values of this map (for proper cleaning from the symbol table when the parent is removed, to avoid leak)
- the `fields` field should not be `pub`. Make it only visible to the `storage` module. Otherwise we cannot trust the keys as valid if any code can change this map.
- add a getter for `fields` so code external to `storage` can read it.

cammarosano: The same applies to `XmlAssetSymbol`. They own the `XmlFieldSymbol`'s stored in `fields`, right?

server/src/core/symbols/module_load.rs
```diff
@@ -23,7 +23,7 @@ use tracing::{error, info};
 impl ModuleSymbol {
 
     pub fn load(mut self, session: &mut SessionInfo, dir_path: &PathBuf) -> Option<Self> {
-        info!("building new module: {:?}", self.path);
+        info!("building new module: {:?}", dir_path.sanitize());
```
cammarosano: like @mmahrouss pointed out, this is a revert of https://github.com/odoo/odoo-ls/commit/39a93c499bdb4ace12789a532cf2815e4dce454a.
Just undo this change, as `self.path` holds the result of `dir_path.sanitize()`, and we don't need to create a new string here.

server/src/core/symbols/module_load.rs
```diff
@@ -36,7 +36,7 @@ impl ModuleSymbol {
         if manifest_file_info.file_info_ast.borrow().indexed_module.is_none() {
             return None;
         }
-        let diags = self.load_manifest(session, &manifest_file_info);
+        let diags = self._load_manifest(session, &manifest_file_info);
```
cammarosano: [conflict-stuff] I suggest removing the addition of `_`

server/src/core/symbols/module_load.rs
```diff
@@ -48,7 +48,7 @@ impl ModuleSymbol {
     }
 
     pub fn load_module_info(module_key: ModuleKey, session: &mut SessionInfo, odoo_addons: NamespaceKey) {
-        let (mut diagnostics, _loaded) = ModuleSymbol::load_depends(module_key, session, odoo_addons);
+        let (mut diagnostics, _loaded) = ModuleSymbol::_load_depends(module_key, session, odoo_addons);
```
cammarosano: [conflict-stuff] I suggest removing the addition of `_`

server/src/core/symbols/module_load.rs
```diff
@@ -64,7 +64,7 @@ impl ModuleSymbol {
 
     /* Load manifest to identify the module characteristics.
     Returns list of od diagnostics to publish in manifest file. */
-    fn load_manifest(&mut self, session: &mut SessionInfo, file_info: &FileInfo) -> Vec<Diagnostic> {
+    fn _load_manifest(&mut self, session: &mut SessionInfo, file_info: &FileInfo) -> Vec<Diagnostic> {
```
cammarosano: [conflict-stuff] I suggest removing the addition of `_`

server/src/core/symbols/module_load.rs
```diff
@@ -338,7 +338,8 @@ impl ModuleSymbol {
 
     /* ensure that all modules indicates in the module dependencies are well loaded.
     Returns list of diagnostics to publish in manifest file */
-    fn load_depends(symbol_key: ModuleKey, session: &mut SessionInfo, odoo_addons: NamespaceKey) -> (Vec<Diagnostic>, Vec<OYarn>) {
+    // @arena: extend a map with a vector??
+    fn _load_depends(symbol_key: ModuleKey, session: &mut SessionInfo, odoo_addons: NamespaceKey) -> (Vec<Diagnostic>, Vec<OYarn>) {
```
cammarosano: [conflict-stuff] I suggest removing the addition of `_`

server/src/core/symbols/module_load.rs
```diff
@@ -348,7 +349,8 @@ impl ModuleSymbol {
         let mut loaded: Vec<OYarn> = vec![];
         let dependencies = module.depends.clone();
         for (depend, range) in dependencies.iter() {
-            if let Some(dependency) = session.sync_odoo.modules.get(depend).and_then(|m| m.upgrade(session.st())) {
+            if let Some(&dependency) = session.sync_odoo.modules.get(depend) {
+                let dependency = dependency.upgrade(session.st()).unwrap();
```
cammarosano: [conflict-stuff] I suggest reverting this change, as this `unwrap` is not necessarily safe. See https://github.com/odoo/odoo-ls/pull/581/changes/698138153ef94463e52b7549565c12c42bf24688

server/src/core/symbols/module_load.rs
```diff
@@ -454,7 +456,7 @@ impl ModuleSymbol {
     }
 
     pub fn load_data(symbol_key: ModuleKey, session: &mut SessionInfo) {
-        let module = &session.st()[symbol_key];
+        let module = &mut session.st_mut()[symbol_key];
```
cammarosano: [conflict-stuff] This does not have to be mutable (revert change).

server/src/core/csv_arch_builder.rs
```diff
@@ -55,10 +49,13 @@ impl CsvArchBuilder {
...
-                        let record = self.extract_record(csv_symbol.into(), model_name.clone(), headers, &result, content);
-                        let Some(record) = record else { continue };
-                        let Some(xml_id) = record.xml_id.as_ref() else { continue };
+                        let record = self.extract_record(session, csv_symbol.into(), model_name.clone(), &result, content);
+                        let Some(record_key) = record else { continue };
+                        let xml_id = {
+                            let record = &session.st()[record_key];
+                            let Some(xml_id) = record.xml_id.clone() else { continue };
+                            xml_id
+                        };
```
cammarosano: Suggestion (simpler, cleaner):

```suggestion
           let Some(xml_id) = session.st()[record_key].xml_id.clone() else { continue };
```

server/src/core/csv_arch_builder.rs
```diff
@@ -86,45 +82,36 @@ impl CsvArchBuilder {
         diagnostics
     }
 
-    fn extract_record(&self, file_symbol: Wk<SymbolKey>, model_name: OYarn, headers: &Vec<OYarn>, record: &StringRecord, content: &String) -> Option<OdooDataRecord> {
+    fn extract_record(&self, session: &mut SessionInfo, file_symbol: SymbolKey, model_name: OYarn, record: &StringRecord, content: &String) -> Option<XmlRecordKey> {
```
cammarosano: `file_symbol` should have type `CsvFileKey` here, as hinted by the call to `unwrap_csv_file_key` below.

So: 

```suggestion
    fn extract_record(&self, session: &mut SessionInfo, file_symbol: CsvFileKey, model_name: OYarn, record: &StringRecord, content: &String) -> Option<XmlRecordKey> {
```
and... (see next comment)

server/src/core/csv_arch_builder.rs
```diff
@@ -86,45 +82,36 @@ impl CsvArchBuilder {
...
+        let record_key = session.st_mut().add_new_xml_record(
+            file_symbol,
+            (model_name, core::ops::Range {
+                start: 0 as usize,
+                end: 1 as usize
+            }),
+            None, //dummy
+            &TextRange::new(TextSize::new(0), TextSize::new(0 as u32)) //dummy
+        );
+        let headers = &session.st()[file_symbol.unwrap_csv_file_key()].headers.clone();
```
cammarosano: ... (continuation of the comment above)
```suggestion
        let headers = &session.st()[file_symbol].headers.clone();
```

server/src/core/csv_arch_builder.rs
```diff
@@ -55,10 +49,13 @@ impl CsvArchBuilder {
             for result in rdr.records() {
                 match result {
                     Ok(result) => {
-                        let headers = &session.st()[csv_symbol].headers;
-                        let record = self.extract_record(csv_symbol.into(), model_name.clone(), headers, &result, content);
-                        let Some(record) = record else { continue };
-                        let Some(xml_id) = record.xml_id.as_ref() else { continue };
+                        let record = self.extract_record(session, csv_symbol.into(), model_name.clone(), &result, content);
```
cammarosano: this is the only call to `extract_record`, which could simply take a `CsvFileKey` for the 2nd argument, and then this `into` is not needed.

server/src/core/data_hooks.rs
```diff
@@ -26,15 +25,23 @@ static record_creation_hooks: Lazy<Vec<RecordCreationHook>> = Lazy::new(|| {
...
+                let field_key: XmlFieldKey = {
+                    let record = &session.st()[record_key];
+                    match record.fields.get("code") {
+                        None => return,
+                        Some(code_field) => match code_field.upgrade(session.st()) {
+                            None => return,
+                            Some(field) => field,
+                        },
+                    }
                 };
```
cammarosano:
```suggestion
let Some(field_key) = session.st()[record_key].fields.get("code")
       .and_then(|code_field| code_field.upgrade(session.st())) else { return };
```

cammarosano: But then see my comment about `XmlFieldKey` being stored as Weak.
I suspect it should not be stored as weak.
In such case, then you don't need the `.and_then(|code_field| code_field.upgrade(session.st()))` above, and the whole hook could become:
```rust
  let Some(language_code) = session.st()[record_key].fields.get("code")
      .and_then(|code_field_key| session.st()[code_field_key].text.clone())
      else { return };
session.sync_odoo.add_language(&language_code source_file);
```

server/src/core/data_hooks.rs
```diff
@@ -45,13 +52,13 @@ static record_creation_hooks: Lazy<Vec<RecordCreationHook>> = Lazy::new(|| {
 pub fn on_record_creation(
     session: &mut SessionInfo,
     source_file: SourceFileKey,
-    record: &OdooDataRecord,
+    record: &XmlRecordKey,
```
cammarosano: `XmlRecordKey` has Copy semantics, so it makes little sense to pass a reference (a pointer) to it instead of the value itself.

```suggestion
    record: XmlRecordKey,
```
(remember to updated the call sites - remove the `&`)

server/src/core/data_hooks.rs
```diff
@@ -45,13 +52,13 @@ static record_creation_hooks: Lazy<Vec<RecordCreationHook>> = Lazy::new(|| {
...
 ) {
-    let model_name = record.model.0.as_str();
+    let model_name = session.st()[*record].model.0.clone();
```
cammarosano:
```suggestion
    let model_name = session.st()[record].model.0.clone();
```

server/src/core/data_hooks.rs
```diff
@@ -45,13 +52,13 @@ static record_creation_hooks: Lazy<Vec<RecordCreationHook>> = Lazy::new(|| {
...
     for hook in record_creation_hooks.iter() {
         // Check if hook applies to this model
-        if hook.models.iter().any(|m| *m == model_name) {
-            (hook.func)(session, source_file, record);
+        if hook.models.iter().any(|m| *m == model_name.as_str()) {
+            (hook.func)(session, source_file, *record);
```
cammarosano:
```suggestion
            (hook.func)(session, source_file, record);
```

server/src/core/data_hooks.rs
```diff
@@ -45,13 +52,13 @@ static record_creation_hooks: Lazy<Vec<RecordCreationHook>> = Lazy::new(|| {
...
-    record: &OdooDataRecord,
+    record: &XmlRecordKey,
 ) {
-    let model_name = record.model.0.as_str();
+    let model_name = session.st()[*record].model.0.clone();
     for hook in record_creation_hooks.iter() {
         // Check if hook applies to this model
-        if hook.models.iter().any(|m| *m == model_name) {
-            (hook.func)(session, source_file, record);
+        if hook.models.iter().any(|m| *m == model_name.as_str()) {
```
cammarosano: you could just flip the equality instead of calling `as_str()`, as Yarn impls PartialEq against `&str`: 
```suggestion
        if hook.models.iter().any(|m|model_name == *m) {
```

server/src/core/odoo.rs
```diff
@@ -1304,13 +1303,13 @@ impl SyncOdoo {
...
-            return vec![];
+            return WeakSet::new();
         }
         let Some(module_key) = module else {
             warn!("Module not found for id: {}", xml_id);
-            return vec![];
+            return WeakSet::new();
         };
-        ModuleSymbol::get_xml_id(session.st(), module_key, id_split.last().unwrap())
+        ModuleSymbol::get_xml_id(&session.st(), module_key, *id_split.last().unwrap()).unwrap_or(WeakSet::new())
```
cammarosano: `session.st()` already returns a borrow/reference to the symbol table. No need to make it a `&&SymbolTable`.
```suggestion
        ModuleSymbol::get_xml_id(session.st(), module_key, *id_split.last().unwrap()).unwrap_or(WeakSet::new())
```

mmahrouss (approved):

server/src/core/symbols/module_load.rs
```diff
@@ -539,19 +541,14 @@ impl ModuleSymbol {
         &self.all_depends
     }
 
+    pub fn insert_xml_id(symbol_table: &mut SymbolTable, target: ModuleKey, xml_id: OYarn, xml_data: XmlId) {
+        symbol_table[target].xml_ids.entry(xml_id).or_insert(WeakSet::new()).insert(xml_data);
```
cammarosano: `WeakSet` implements `Default`, so you could simply:
```suggestion
        symbol_table[target].xml_ids.entry(xml_id).or_default().insert(xml_data);
```
It's also more efficient, at least in theory: it will not create an empty WeakSet unless it's needed, while `WeakSet::new()` gets always evaluated.

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -117,6 +118,60 @@ impl SymbolTable {
         xml_file_key
     }
 
+    pub fn add_new_xml_record(&mut self, parent: SymbolKey, model: (OYarn, Range<usize>) , xml_id: Option<OYarn>, range: &TextRange) -> XmlRecordKey {
+        let is_external = self.is_external(parent.into());
+        let xml_record_sym = XmlRecordSymbol::new(
+            model,
+            xml_id,
+            range.clone(),
+            parent.into(),
```
cammarosano: `parent.into()` currently converts a `SymbolKey` into a `SymbolKey`
```suggestion
            parent,
```

cammarosano: to catch similar stuff:
`cargo clippy 2>&1 | grep -E "useless conversion to the same type" -A 10`

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -199,6 +253,38 @@ impl SymbolTable {
         ns.directories[best].module_symbols.insert(oyarn!("{}", name), file)
     }
 
+    fn add_field_to_xml_record(&mut self, parent: SymbolKey, field: XmlFieldKey, name: &str) {
+        match parent {
+            SymbolKey::XmlRecord(r) => {
+                let xml_record = &mut self.xml_records[r];
+                xml_record.fields.insert(oyarn!("{}", name), Wk::from(field));
```
cammarosano: This function is a helper for `add_new_xml_field`, which creates (and inserts on the table) a new `XmlFieldSymbol`.
And this is the place where this new symbol will "live": in the `fields` maps in `XmlRecord` or `XmlAsset`. So this should be a strong key, not a weak. It's a parent-child relationship.

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -367,7 +465,13 @@ impl SymbolTable {
             SymbolKey::Class(c) => { self.classes[c].symbols.remove(&child_name); },
             SymbolKey::Function(f) => { self.functions[f].symbols.remove(&child_name); },
             SymbolKey::Variable(_) => { panic!("A variable cannot be a parent") },
-            SymbolKey::XmlFile(_) => { panic!("An XML file symbol cannot be a parent") },
+            SymbolKey::XmlFile(f) => { self.xml_files[f].symbols.remove(&child.as_xml_data_key().expect("Content of xmlfile should be an xml_data_key")); },
+            SymbolKey::XmlRecord(_) => { panic!("An XML record cannot be a parent") },
```
cammarosano: It *should* be the parent of xml fields!
So unlinking a `XmlFieldKey` from a `XmlField` parent should actually remove it from the `fields` map.

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -367,7 +465,13 @@ impl SymbolTable {
...
+            SymbolKey::XmlField(_) => { panic!("An XML field cannot be a parent") },
+            SymbolKey::XmlAsset(_) => { panic!("An XML asset cannot be a parent") },
```
cammarosano: Same. It should be the parent of `XmlFieldKey`s.

server/src/core/symbols/storage/xml/xml_file_symbol.rs
```diff
@@ -18,8 +19,8 @@ pub struct XmlFileSymbol {
     pub validation_status: BuildStatus,
     pub not_found_paths: Vec<(BuildSteps, Vec<OYarn>)>,
     pub not_found_models: HashMap<OYarn, BuildSteps>,
-    pub xml_ids: HashMap<OYarn, Vec<OdooData>>,
-    pub (super) in_workspace: bool,
+    pub symbols: HashSet<XmlDataKey>,
```
cammarosano: This should not be public (only the `storage` module can make additions to it).
Then just a add a getter so that outside code can read it.

server/src/core/xml_arch_builder.rs
```diff
@@ -83,21 +80,20 @@ impl XmlArchBuilder {
...
         let xml_ids = SyncOdoo::get_xml_ids(session, self.xml_symbol.into(), xml_id, &attr.range(), diagnostics);
         let mut res = vec![];
-        for data in xml_ids.iter() {
-            if let OdooData::RECORD(r) = data {
-                if r.model.0 == "res.groups" {
+        for data in xml_ids.iter_valid(session.st()) {
+            if let XmlId::XmlRecord(r) = data {
+                let record = &session.st()[r];
+                if record.model.0 == "res.groups" {
                     res.push(data.clone());
```
cammarosano: Not really from your commit, but now that I saw this...
XmlId has Copy semantics! (no clone needed)
```suggestion
                    res.push(data);
```

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -169,12 +170,12 @@ impl XmlArchBuilder {
...
         }
-        let data = OdooData::MENUITEM(XmlDataMenuItem {
-            file_symbol: self.xml_symbol.into(),
-            xml_id: found_id.clone().map(|id| oyarn!("{}", id)),
-            range: node.range().clone()
-        });
-        self.on_operation_creation(session, found_id, node, data, diagnostics);
+        let data = session.st_mut().add_new_xml_menuitem(
+            self.xml_symbol,
+            found_id.clone().map(|id| oyarn!("{}", id)),
```
cammarosano: Since you already own the String after the `.clone`, you can skip the whole format machinery:
```suggestion
            found_id.clone().map(OYarn::from),
```

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -629,6 +631,10 @@ impl XmlArchBuilder {
                 });
             }
         }
+        let asset = session.st_mut().add_new_xml_asset(
+            self.xml_symbol,
+            found_id.clone().map(|id| oyarn!("{}", id)),
```
cammarosano: Same as before (you already own the String after `clone`)
```suggestion
            found_id.clone().map(OYarn::from),
```

server/src/core/xml_validation.rs
```diff
@@ -82,63 +76,63 @@ impl XmlValidator {
         file_info.borrow_mut().publish_diagnostics(session);
     }
 
-    fn validate_data(&mut self, session: &mut SessionInfo, data: &OdooData, diagnostics: &mut Vec<Diagnostic>, dependencies: &mut Vec<SourceFileKey>, model_dependencies: &mut Vec<Rc<RefCell<Model>>>, missing_model_dependencies: &mut HashSet<OYarn>) {
-        let Some(_) = data.get_xml_file_symbol(&session.sync_odoo.symbol_table) else {
+    fn validate_data(&mut self, session: &mut SessionInfo, data: &XmlDataKey, diagnostics: &mut Vec<Diagnostic>, dependencies: &mut Vec<SourceFileKey>, model_dependencies: &mut Vec<Rc<RefCell<Model>>>, missing_model_dependencies: &mut HashSet<OYarn>) {
```
cammarosano: keys have Copy semantics! It's not idiomatic to pass references/pointers to copiable data.
I't like passing `&5` to a function instead of just `5`. 
And you add an unnecessary indirection layer (needing to dereference the pointer to read data).
I suggest you remove this `&` and all the `*`s that came as a consequence.

```suggestion
    fn validate_data(&mut self, session: &mut SessionInfo, data: XmlDataKey, diagnostics: &mut Vec<Diagnostic>, dependencies: &mut Vec<SourceFileKey>, model_dependencies: &mut Vec<Rc<RefCell<Model>>>, missing_model_dependencies: &mut HashSet<OYarn>) {
```

server/src/core/xml_validation.rs
```diff
@@ -82,63 +76,63 @@ impl XmlValidator {
...
+            XmlDataKey::TEMPLATE(xml_data_template) => self.validate_template(session, xml_data_template, diagnostics, dependencies, model_dependencies, missing_model_dependencies),
+            XmlDataKey::DELETE(xml_data_delete) => self.validate_delete(session, xml_data_delete, diagnostics, dependencies, model_dependencies, missing_model_dependencies),
+            XmlDataKey::ASSET(xml_data_asset) => self.validate_asset(session, xml_data_asset, diagnostics, dependencies, model_dependencies, missing_model_dependencies),
         }
     }
-    fn validate_record(&mut self, session: &mut SessionInfo, xml_data_record: &OdooDataRecord, diagnostics: &mut Vec<Diagnostic>, dependencies: &mut Vec<SourceFileKey>, model_dependencies: &mut Vec<Rc<RefCell<Model>>>, missing_model_dependencies: &mut HashSet<OYarn>) {
-        let maybe_model = session.sync_odoo.models.get(&xml_data_record.model.0).cloned();
+    fn validate_record(&mut self, session: &mut SessionInfo, xml_data_record: &XmlRecordKey, diagnostics: &mut Vec<Diagnostic>, dependencies: &mut Vec<SourceFileKey>, model_dependencies: &mut Vec<Rc<RefCell<Model>>>, missing_model_dependencies: &mut HashSet<OYarn>) {
```
cammarosano: same (key impls Copy)

server/src/core/xml_validation.rs
```diff
@@ -82,63 +76,63 @@ impl XmlValidator {
...
             self.fields_cache.insert(model_name.clone(), all_fields);
         }
-        let all_fields = self.fields_cache.get(model_name).unwrap();
-        self.validate_fields(session, xml_data_record, all_fields, diagnostics, missing_model_dependencies);
+        let all_fields = self.fields_cache.get(&model_name).unwrap();
+        self.validate_fields(session, &xml_data_record, all_fields, diagnostics, missing_model_dependencies);
     }
 
-    fn validate_fields(&self, session: &mut SessionInfo, xml_data_record: &OdooDataRecord, all_fields: &HashMap<OYarn, Vec<(SymbolKey, Option<OYarn>)>>, diagnostics: &mut Vec<Diagnostic>, missing_model_dependencies: &mut HashSet<OYarn>) {
+    fn validate_fields(&self, session: &mut SessionInfo, xml_data_record: &XmlRecordKey, all_fields: &HashMap<OYarn, Vec<(SymbolKey, Option<OYarn>)>>, diagnostics: &mut Vec<Diagnostic>, missing_model_dependencies: &mut HashSet<OYarn>) {
```
cammarosano: same (`XmlRecordKey` impls `Copy`)

server/src/core/xml_validation.rs
```diff
@@ -284,19 +284,19 @@ impl XmlValidator {
         // }
     }
 
-    fn validate_menu_item(&self, _session: &mut SessionInfo, _xml_data_menu_item: &XmlDataMenuItem, _diagnostics: &mut Vec<Diagnostic>, _dependencies: &mut Vec<SourceFileKey>, _model_dependencies: &mut Vec<Rc<RefCell<Model>>>, _missing_model_dependencies: &mut HashSet<OYarn>) {
+    fn validate_menu_item(&self, _session: &mut SessionInfo, _xml_data_menu_item: &XmlMenuItemKey, _diagnostics: &mut Vec<Diagnostic>, _dependencies: &mut Vec<SourceFileKey>, _model_dependencies: &mut Vec<Rc<RefCell<Model>>>, _missing_model_dependencies: &mut HashSet<OYarn>) {
```
cammarosano: Same for all these signature. These keys impl `Copy`, there's no point in passing a reference to them.

server/src/features/csv_ast_utils.rs
```diff
@@ -170,24 +170,28 @@ impl CsvAstUtils {
...
-                        xml_id: Some(field_data),
-                        fields: vec![],
-                        range: core::ops::Range{
-                            start,
-                            end,
-                        },
-                    })),
-                    origin_selection_range: None,
-                })
+                for data in session.st()[csv_symbol.unwrap_csv_file_key()].symbols.iter() {
```
cammarosano: this is an indicator that `csv_symbol` should have type `CsvFileKey` instead of the more generic `SourceFileKey` (as its name already implies). Changing it here would propagate the change through its callers. So maybe just leave a "todo" here for later?

cammarosano (changes requested):
Nice! The symbol family is growing!
Other that style/minor improvements, the really important part is the parent-child relationships (strong keys), the visibility of the children maps/sets (should not be pub) and the removal of these children on key remove and unlink_from_parent.

server/src/core/xml_arch_builder_rng_validation.rs
fda-odoo: Yes, it's dead code, that was there to be added to record after the loop, but load_field is handling it now

server/tests/test_get_symbol.rs
```diff
@@ -545,7 +546,7 @@ fn test_csv_quoted_commas() {
     // quoted field, the record must still be registered. With quoting=false
     // (old bug) the interior comma caused a field-count error → record skipped.
     assert!(
-        session.st()[module].xml_id_locations.contains_key("state_comma_1"),
+        session.st()[module].xml_ids.contains_key(&Sy!("state_comma_1")),
```
cammarosano: the mysterious addition of `Sy!` 😄  (not necessary)

server/tests/test_get_symbol.rs
```diff
@@ -622,7 +623,7 @@ fn test_csv_ranges_unquoted_lf(session: &mut SessionInfo) {
         .upgrade(session.st())
         .unwrap();
     assert!(
-        session.st()[module].xml_id_locations.contains_key("state_unquoted_1"),
+        session.st()[module].xml_ids.contains_key(&Sy!("state_unquoted_1")),
```
cammarosano: same

server/tests/test_get_symbol.rs
```diff
@@ -689,7 +690,7 @@ fn test_csv_ranges_quoted_lf(session: &mut SessionInfo) {
         .upgrade(session.st())
         .unwrap();
     assert!(
-        session.st()[module].xml_id_locations.contains_key("state_quoted_1"),
+        session.st()[module].xml_ids.contains_key(&Sy!("state_quoted_1")),
```
cammarosano: same

server/tests/test_get_symbol.rs
```diff
@@ -751,7 +752,7 @@ fn test_csv_ranges_unquoted_crlf(session: &mut SessionInfo) {
         .upgrade(session.st())
         .unwrap();
     assert!(
-        session.st()[module].xml_id_locations.contains_key("state_unquoted_crlf_1"),
+        session.st()[module].xml_ids.contains_key(&Sy!("state_unquoted_crlf_1")),
```
cammarosano: same

server/tests/test_get_symbol.rs
```diff
@@ -817,7 +818,7 @@ fn test_csv_ranges_quoted_crlf(session: &mut SessionInfo) {
         .upgrade(session.st())
         .unwrap();
     assert!(
-        session.st()[module].xml_id_locations.contains_key("state_quoted_crlf_1"),
+        session.st()[module].xml_ids.contains_key(&Sy!("state_quoted_crlf_1")),
```
cammarosano: same

server/src/features/xml_ast_utils.rs
```diff
@@ -120,7 +102,7 @@ impl XmlAstUtils {
                             if missing_dep.is_none() {
                                 let content = session.sync_odoo.symbol_table.get_content_symbol(class_key.into(), attr.value(), u32::MAX);
                                 for symbol in content.symbols.iter() {
-                                    results.0.push(XmlAstResult::SYMBOL(*symbol));
+                                    results.0.push(symbol.clone());
```
cammarosano:
```suggestion
                                    results.0.push(*symbol);
```

Or, because keys have Copy semantics, you could do simply:
```rust
                                for symbol in content.symbols {
                                    results.0.push(symbol);
                                }
```
(no `iter`, and you simply iterate over `SymbolKey`s, no `&`)

cammarosano (commented):
A few extra minor suggestions

server/src/core/symbols/storage/csv_file_symbol.rs
```diff
@@ -15,7 +16,7 @@ pub struct CsvFileSymbol {
     pub validation_status: BuildStatus,
     pub not_found_paths: Vec<(BuildSteps, Vec<OYarn>)>,
     pub (super) in_workspace: bool,
-    pub xml_ids: HashMap<OYarn, Vec<OdooData>>,
+    pub symbols: HashSet<XmlDataKey>,
```
cammarosano:
```suggestion
    pub (super) symbols: HashSet<XmlDataKey>,
```
I did not see this one in my first review, sorry.
It should be visible only to the storage (super) module, then add a getter so that code in csv_ast_utils can read it.

server/src/core/symbols/storage/csv_file_symbol.rs
```diff
@@ -60,7 +61,7 @@ impl CsvFileSymbol {
 
     /// no child symbols
```
cammarosano: stale comment

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -199,6 +253,38 @@ impl SymbolTable {
...
+                xml_record.fields.insert(oyarn!("{}", name), field);
+            },
+            SymbolKey::XmlAsset(k) => {
+                let xml_asset = &mut self.xml_assets[k];
+                xml_asset.fields.insert(oyarn!("{}", name), field);
+            }
+            _ => {
+                panic!("Impossible to add an XmlFieldKey to a {}", parent.typ());
+            }
+        }
```
cammarosano, line 273: If a XmlFieldKey ever gets replaced (due to insertion with the same name), that key becomes unreacheable, and effectively removed from the symbol tree.
Which is correct.
But then we have to clear that symbol from the symbol storage, otherwise it's a memory leak.
At this point, I'm not sure that's possible to happen, but we have something similar in `add_to_parent_module_symbols`, in which it did happen.
So just collect the replaced key (if any), and call `self.remove()` on it:

```suggestion
        let replaced_key = match parent {
            SymbolKey::XmlRecord(r) => {
                let xml_record = &mut self.xml_records[r];
                xml_record.fields.insert(oyarn!("{}", name), field)
            },
            SymbolKey::XmlAsset(k) => {
                let xml_asset = &mut self.xml_assets[k];
                xml_asset.fields.insert(oyarn!("{}", name), field)
            }
            _ => {
                panic!("Impossible to add an XmlFieldKey to a {}", parent.typ());
            }
        };
        if let Some(replaced_key) = replaced_key {
            self.remove(replaced_key.into());
        }
```

server/src/core/symbols/storage/xml/xml_record_symbol.rs
```diff
@@ -0,0 +1,43 @@
...
+use ruff_text_size::TextRange;
+
+use crate::{constants::OYarn, core::symbols::symbol_keys::{SymbolKey, XmlFieldKey}};
+
+#[derive(Debug)]
+pub struct XmlRecordSymbol {
+    pub is_external: bool,
+    pub model: (OYarn, Range<usize>),
+    pub xml_id: Option<OYarn>,
+    pub fields: HashMap<OYarn, XmlFieldKey>,
```
cammarosano: you forgot to make this one visible only for `storage` crate

server/src/features/references.rs
```diff
@@ -91,10 +107,10 @@ impl ReferenceFeature {
                         }
                     }
                     //If the symbol is a model or a field, browse model dependents too
-                    let class_model_to_check = if target_symbol.typ() == SymType::CLASS {
-                        Some(target_symbol)
-                    } else if SymbolTable::is_field(session, target_symbol) {
-                        session.st().get_in_parents(target_symbol, &[SymType::CLASS], true)
+                    let class_model_to_check = if definition.source.typ() == SymType::CLASS {
+                        Some(definition.source.clone())
```
cammarosano: No need to `.clone()`
```suggestion
                        Some(definition.source)
```

server/src/features/workspace_symbols.rs
```diff
@@ -80,24 +80,25 @@ impl WorkspaceSymbolFeature {
...
+                        let xml_file_symbol = session.st().get_file(data.into());
+                        if let Some(SourceFileKey::XmlFile(xml_file_key)) = xml_file_symbol {
                             let xml_file = &session.st()[xml_file_key];
                             let name = xml_file.name.to_string();
                             let path = xml_file.path.clone();
-                            let range = data.get_range();
-                            let text_range = TextRange::new(TextSize::new(range.start as u32), TextSize::new(range.end as u32));
-                            WorkspaceSymbolFeature::add_symbol_to_results(session, xml_file_key.into(), &xml_name, &path, Some(name), Some(&text_range), can_resolve_location_range, results);
+                            let data_range = session.st().range(data.into()).clone();
+                            res_to_add.push((xml_file_key, xml_name.clone(), path.clone(), name.clone(), data_range));
```
cammarosano: `name` and `path` are owned Strings you don't need anymore after this call, so you can just move them instead of cloning again.
```suggestion
                            res_to_add.push((xml_file_key, xml_name.clone(), path, name, data_range));
```

cammarosano (changes requested):
Nice! You're the master of conflict resolution! 😄 
Just a few things to change, and we're ready to 🚀.

cammarosano (approved):
Good work, thanks for the patience.

PR #589, opened 2026-04-23
[FIX] Correctly get file path URIs

mmahrouss:
- Tracked folders being required is now enforced
- Use FileMgr::try_pathname2uri instead of Uri::from_str, that is because from_str just does a naïve conversion, but workspace folders expects having a file URI


server/src/cli_backend.rs
mmahrouss, line 1: Make sure you put the diff at "Hide Whitespace" to avoid seeing the indentation diff

fda-odoo (approved):

PR #593, opened 2026-05-05
[FIX] Correctly handle files that are opened but not created

mmahrouss:
This was discovered from a crash report, where a module is moved/copied into odoo/addons, and that was causing a panic. Because a custom EP was getting wrongfully created.

This PR makes sure new files that are seen in did open, go through the did open pipe. It also checks whether the new file has a parent module that is not loaded yet, and loads it

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -137,7 +137,6 @@ pub struct SyncOdoo {
     pub import_cache: Option<ImportCache>,
     pub capabilities: lsp_types::ClientCapabilities,
     pub encoding: PositionEncoding,
-    pub opened_files: Vec<String>,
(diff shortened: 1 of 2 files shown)


server/src/core/odoo.rs
```diff
@@ -137,7 +137,7 @@ pub struct SyncOdoo {
     pub import_cache: Option<ImportCache>,
     pub capabilities: lsp_types::ClientCapabilities,
     pub encoding: PositionEncoding,
-    pub opened_files: Vec<String>,
```
mmahrouss, line 140: `opened_files` was never read, so I replaced it with `created_files` that actually serves a purpose in this PR

mmahrouss: and a vector to avoid iterating over the whole vector everytime when we call retain

mmahrouss:
This was the wrong solution, it was due to the server not requesting to watch for FS create/delete. Will create a new PR, and keep this one for reference

PR #594, opened 2026-05-06
[FIX] Fix panic on itering symbols on Disk Dir

mmahrouss:
This happens when you try to import from a directory in a custom entry point. that should be a namespace/package but is not resolved

A task should follow to fix import issues in custom EntryPoints

fixes error 2100

### DIFF
--- a/server/src/core/python_arch_builder.rs
+++ b/server/src/core/python_arch_builder.rs
@@ -222,20 +222,19 @@ impl PythonArchBuilder {
                     }
                     let mut dep_to_add = vec![];
                     let sym_type = import_symbol.borrow().typ();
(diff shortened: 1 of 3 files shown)


fda-odoo:
that's solving the crash, but not fixing the bug itself, that prevent importing from custom EP
let's keep that for 1.3.2, but we should check at that in future version

fda-odoo (approved):

mmahrouss:
> that's solving the crash, but not fixing the bug itself, that prevent importing from custom EP let's keep that for 1.3.2, but we should check at that in future version

Yes I agree, I opened a task for future versions:)

Issue #508, opened 2025-12-10
`freezegun` and packages with inlined types do not autocomplete


andg-odoo:
**Your setup**
Version: 1.1.2
Operating System: Ubuntu 24.04
IDE and/or Integration tool (for example: Vscode - official extension): Neovim 0.12

**Describe the bug**
Packages that don't have their typing within the typeshed module don't have import autocomplete. 

This is because `freezegun` and other packages are removed from python/typeshed#5842 when the package added it's own stubs.

**Expected behavior**
For python packages that contains inlined stubs, is it possible to directly detect them? Or do we have to go and put each folder from the python environment in the `additional_stubs` key of the config.
<img width="834" height="56" alt="Image" src="https://github.com/user-attachments/assets/db02ccc1-de32-4211-a5d4-f935c0f3209f" />

fda-odoo:
@andg-odoo I'm looking into imports and checking your issue.
For packages that are not included in typeshed, we will not provide stubs (except if there is another official source for stubs)
In the case of freezegun, installing the package in your python environment seems to work, and I have autocompletion when importing it.
You should not put each folder of your venv into `additional_stubs` but rather set the `python_path` value to the venv one you are using. Odoo-ls will then take the virtualenv setup for dependencies
for example:
```toml
[[config]]
python_path = "/home/me/Documents/.venv/bin/python"
```

fda-odoo:
(tested on future 1.3.2)

andg-odoo:
Hmm interesting, I wonder if this is instead a pyenv issue? Here is my `odools.toml`:
```toml
[[config]]
name = "Custom Setup"
odoo_path = "/home/andg/Dev/src/odoo"
addons_paths = ["/home/andg/Dev/src/enterprise"]
additional_stubs = ["/home/andg/.local/share/nvim/odoo/typeshed/stubs"]
python_path = "/home/andg/.pyenv/shims/python3"
stdlib = "/home/andg/.local/share/nvim/odoo/typeshed/stdlib"
```

<img width="723" height="246" alt="Image" src="https://github.com/user-attachments/assets/dd953842-2c82-4484-9abc-d483b472179a" />

It could be with the symlink similar to the discussion with DYKI

I've just pulled alpha and still have the issue for markupsafe and freezegun.

EDIT: I've swapped my python_path to: `python_path = "/home/andg/.pyenv/versions/odoo/bin/python3"` and it still has the same errors.

andg-odoo:
[...]

PR #595, opened 2026-05-07
[FIX] Only trim .py[i] from suffixes

mmahrouss:
See https://github.com/odoo/odoo-ls/issues/508#issuecomment-4393690936

### DIFF
--- a/server/src/utils.rs
+++ b/server/src/utils.rs
@@ -2,6 +2,7 @@ use crate::core::file_mgr::legacy_unc_paths;
 use path_slash::{PathBufExt, PathExt};
 use regex::Regex;
 use ruff_text_size::TextSize;
+use std::ffi::OsStr;
 use std::process::Command;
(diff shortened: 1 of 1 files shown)


server/src/utils.rs
```diff
@@ -182,14 +183,29 @@ impl PathSanitizer for PathBuf {
...
+        {
+            self.with_extension("")
+        } else {
+            self.clone()
+        };
+        // Drop __init__ or __manifest__ if they are the file name
+        // Use the parent directory path
+        if matches!(
+            modified_path.file_name().and_then(OsStr::to_str),
+            Some("__init__") | Some("__manifest__")
```
fda-odoo: it will remove any __init__ and __manifest__ files, independently of the extension

mmahrouss: Yes you are right, will fix it now:) - sent from vscode:D

fda-odoo (approved):

PR #597, opened 2026-05-07
[IMP] OLS03025: discourage mutable variables outside of models.

fda-odoo:
Mutables variables could lead to unexpected behaviour in a multi-worker context. This warning aims to discourage this usage


fda-odoo:
Not ready to be merged yet.
We still need to analyze the results and see if we want to keep both list and dict warning, or none of them.
This branch actually stay for analysis

PR #599, opened 2026-05-08
Fix file watched capabilities and data file checks

mmahrouss:


fda-odoo (approved):

fda-odoo:
Great :) that's way shorter indeed :D

PR #600, opened 2026-05-13
[IMP] Add XML model records to models

mmahrouss:
This adds xml models and fields in the model struct. It is also handled in hover, definition, ... .
Also includes tests

### DIFF
--- a/server/src/core/csv_validation.rs
+++ b/server/src/core/csv_validation.rs
@@ -6,9 +6,7 @@ use tracing::info;
 
 use crate::{
     Sy, constants::{BuildStatus, BuildSteps, DEBUG_STEPS, OYarn}, core::{
-        diagnostics::{DiagnosticCode, create_diagnostic},
-        file_mgr::FileInfo,
-        symbols::{storage::SymbolTable, symbol_keys::{CsvFileKey, ModuleKey}},
+        diagnostics::{DiagnosticCode, create_diagnostic}, evaluation_utils::DeepFieldEvalWalker, file_mgr::FileInfo, symbols::{SymbolTable, symbol_keys::{CsvFileKey, ModuleKey}}
     }, features::csv_ast_utils::CsvFieldIter, threads::SessionInfo
 };
 use std::{cell::RefCell, rc::Rc};
@@ -58,15 +56,27 @@ impl CsvValidator {
             self.finalize_validation(session, csv_symbol, &file_info, diagnostics);
             return;
         };
-        let model_main_sym = model.borrow().get_main_symbols(session, Some(csv_module));
-        let Some(&model_main_sym) = model_main_sym.get(0) else { return;};
+        let model_main_class_sym = {
+            let model_ref = model.borrow();
+            let mut model_main_sym = model_ref.get_main_symbols(session, Some(csv_module));
+            let Some(model_main_sym) = model_main_sym.find_map(|s| s.as_class_key()) else {
+                return;
+            };
+            model_main_sym
+        };
         if rdr.has_headers() && let Ok(header) = rdr.headers() {
             let mut header_is_xml = vec![false; header.len()];
             for (idx, (start, end, h)) in CsvFieldIter::new(header, &data).unwrap().enumerate() {
-                let mut header_elts = h.splitn(2, [':', '/']).collect::<Vec<_>>();
-                header_elts[0] = header_elts[0].split("@").next().unwrap(); //remove translation if exists
-                let member_sym = SymbolTable::get_member_symbol(session, model_main_sym.into(), header_elts[0], Some(csv_module), false, true, false, true, false);
-                if member_sym.0.is_empty() {
+                let header_elts = h.splitn(2, [':', '/']).collect::<Vec<_>>();
+                let field_name = header_elts[0].split("@").next().unwrap(); //remove translation if exists
+                let mut deep_field_walker =
+                    DeepFieldEvalWalker::new(model_main_class_sym.into(), Some(csv_module));
+                let member_symbols = deep_field_walker.get_model_fields(
+                    session,
+                    model_main_class_sym.into(),
+                    field_name,
+                );
+                if member_symbols.is_empty() {
                     header_is_xml[idx] = false;
                     if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05057, &[h, model_name]) {
                         diagnostics.push(Diagnostic {
@@ -75,18 +85,9 @@ impl CsvValidator {
                         });
                     }
                 } else {
-                    let mut is_relational = false;
-                    for sym in member_sym.0.iter() {
-                        if SymbolTable::is_specific_field(session, *sym, &["Many2one", "One2many", "Many2many"]) &&
-                            header_elts.len() == 2 && header_elts[1] == "id"
-                        {
-                            //on relational fields, the header must contain :id to search for xml_id.
-                            //Else, it will use _rec_name and _rec_name_search to find the right record, which we are not validating
-                            is_relational = true;
-                            break;
-                        }
-                    }
-                    header_is_xml[idx] = is_relational;
+                    header_is_xml[idx] = header_elts.len() == 2
+                        && header_elts[1] == "id"
+                        && deep_field_walker.last_field_is_relational(session);
                 }
             }
             if session.st()[csv_symbol].headers.contains(&Sy!("id")) {
@@ -107,8 +108,8 @@ impl CsvValidator {
     fn validate_record(&self, session: &mut SessionInfo, csv_module: ModuleKey, headers_is_xml: &Vec<bool>, record: &StringRecord, diagnostics: &mut Vec<Diagnostic>, data: &str) {
         let Some(field_iter) = CsvFieldIter::new(record, data) else { return; };
         for (idx, (start, end, field)) in field_iter.enumerate() {
-            let Some(should_be_xml_id) = headers_is_xml.get(idx) else { break;};
-            if *should_be_xml_id {
+            let Some(&should_be_xml_id) = headers_is_xml.get(idx) else { break;};
+            if should_be_xml_id {
                 //check that field_data is a valid xml id
                 let id_split = field.split(".").collect::<Vec<&str>>();
                 let mut module_name = session.st().name(csv_module).as_str();
--- a/server/src/core/evaluation_utils.rs
+++ b/server/src/core/evaluation_utils.rs
@@ -1,0 +1,153 @@
+use crate::{
+    core::symbols::{
+        SymbolTable, VariableSymbol,
+        storage::xml::xml_field_symbol::XmlFieldName,
+        symbol_keys::{ModuleKey, SymbolKey},
+    },
+    threads::SessionInfo,
+};
+
+/// Helper struct to walk through relational fields of a model in order to find the symbol of a nested field
+/// It follows the chain of relational fields using field string names
+/// It is used by alternating the call between get_model_symbol and get_model_fields,
+/// first to find the model symbol of the next field, then to find the field symbols of the next field.
+/// Gives the possibility to process the fields or the intermediate models at each step.
+/// last_field_is_relational can be used after get_model_fields to test whether the fetched
+/// fields are relational without resolving their comodel.
+pub struct DeepFieldEvalWalker {
+    base_object: SymbolKey,
+    from_module: Option<ModuleKey>,
+    prev_field_symbols: Option<Vec<SymbolKey>>, // field_symbols from the previous iteration, used to find the next base object
+}
+
+impl DeepFieldEvalWalker {
+    pub fn new(base_object: SymbolKey, from_module: Option<ModuleKey>) -> Self {
+        Self {
+            base_object,
+            from_module,
+            prev_field_symbols: None,
+        }
+    }
+
+    /// Find the next model symbol by looking for relational fields in the current field symbols
+    /// Initial step of the walker
+    /// Returns None if the previous step yielded a non relational field
+    pub fn get_model_symbol(&mut self, session: &mut SessionInfo) -> Option<SymbolKey> {
+        let field_symbols = match &self.prev_field_symbols {
+            Some(field_symbols) => field_symbols,
+            None => return Some(self.base_object), // First iteration, use the initial base object
+        };
+        for field_symbol_key in field_symbols.iter() {
+            let model_symbols = match field_symbol_key {
+                SymbolKey::XmlRecord(key) => {
+                    let Some(ttype) =
+                        session.st()[*key].get_field_text(XmlFieldName::Type, session.st())
+                    else {
+                        continue;
+                    };
+                    if !["many2one", "many2many", "one2many"].contains(&ttype.as_str()) {
+                        continue;
+                    }
+                    let Some(relation) =
+                        session.st()[*key].get_field_text(XmlFieldName::Relation, session.st())
+                    else {
+                        continue;
+                    };
+                    let Some(related_models) = session.sync_odoo.models.get(relation.as_str())
+                    else {
+                        continue;
+                    };
+                    related_models
+                        .borrow()
+                        .get_main_symbols(session, self.from_module)
+                        .collect::<Vec<_>>()
+                }
+                SymbolKey::Variable(v) => {
+                    if SymbolTable::is_specific_field(
+                        session,
+                        *field_symbol_key,
+                        &["Many2one", "One2many", "Many2many"],
+                    ) {
+                        VariableSymbol::get_relational_model(*v, session, self.from_module)
+                    } else {
+                        vec![]
+                    }
+                }
+                _ => vec![],
+            };
+            //only handle it if there is only one main symbol for this model
+            if model_symbols.len() == 1 {
+                return Some(model_symbols[0].into()); // TODO; handle multuple values
+            }
+        }
+        None
+    }
+
+    /// Returns true if any field symbol from the previous get_model_fields call is relational.
+    /// Cheaper than get_model_symbol: it does not resolve the comodel, and it is also correct
+    /// for the multi-comodel case where get_model_symbol would return None.
+    pub fn last_field_is_relational(&self, session: &mut SessionInfo) -> bool {
+        let Some(field_symbols) = &self.prev_field_symbols else {
+            return false;
+        };
+        field_symbols.iter().any(|field_symbol_key| match field_symbol_key {
+            SymbolKey::XmlRecord(key) => session.st()[*key]
+                .get_field_text(XmlFieldName::Type, session.st())
+                .is_some_and(|ttype| ["many2one", "many2many", "one2many"].contains(&ttype.as_str())),
+            SymbolKey::Variable(_) => SymbolTable::is_specific_field(
+                session,
+                *field_symbol_key,
+                &["Many2one", "One2many", "Many2many"],
+            ),
+            _ => false,
+        })
+    }
+
+    /// Get model fields with the given name
+    /// Returns an empty vec if the base object is not a model or if there are no fields with the given name
+    pub fn get_model_fields(
+        &mut self,
+        session: &mut SessionInfo,
+        base_object: SymbolKey,
+        name: &str,
+    ) -> Vec<SymbolKey> {
+        let symbols = match base_object {
+            SymbolKey::XmlRecord(xml_record_key) => {
+                if let Some(model) = SymbolTable::get_xml_defined_model(session, xml_record_key) {
+                    model
+                        .borrow()
+                        .get_xml_model_field_symbols(session.st(), self.from_module)
+                        .filter_map(|f_key| {
+                            let field_name = session.st()[f_key]
+                                .get_field_text(XmlFieldName::Name, session.st())?;
+                            if field_name == name {
+                                Some(SymbolKey::from(f_key))
+                            } else {
+                                None
+                            }
+                        })
+                        .collect::<Vec<_>>()
+                } else {
+                    vec![]
+                }
+            }
+            SymbolKey::Class(_) => {
+                let (symbols, _diagnostics) = SymbolTable::get_member_symbol(
+                    session,
+                    base_object,
+                    name,
+                    self.from_module,
+                    false,
+                    true,
+                    false,
+                    true,
+                    false,
+                );
+                symbols
+            }
+            _ => Vec::new(),
+        };
+        self.prev_field_symbols = Some(symbols.clone());
+        symbols
+    }
+}
--- a/server/src/core/model.rs
+++ b/server/src/core/model.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use lsp_types::MessageType;
 use std::cell::RefCell;
 use crate::utils::HashMap;
@@ -6,7 +7,10 @@ use std::collections::VecDeque;
 use std::rc::Rc;
 
 use crate::constants::OYarn;
+use crate::core::symbols::symbol_keys::ModelSymbolKey;
 use crate::core::symbols::symbol_keys::SourceFileKey;
+use crate::core::symbols::symbol_keys::SymbolKey;
+use crate::core::symbols::symbol_keys::XmlRecordKey;
 use crate::core::symbols::symbol_keys::{ClassKey, ModuleKey};
 use crate::core::symbols::storage::SymbolTable;
 use crate::threads::SessionInfo;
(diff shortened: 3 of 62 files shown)


.vscode/launch.json
```diff
@@ -38,7 +38,7 @@
             "cwd": "${workspaceFolder}/server",
             "externalConsole": false,
             "MIMode": "gdb",
-            "miDebuggerPath": "rust-gdb",
```
mmahrouss, line 41: Taken from https://github.com/odoo/odoo-ls/pull/601, so we should merge it first to preserve the ownership

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -711,4 +715,102 @@ impl XmlArchBuilder {
         self.on_operation_creation(session, found_id, node, asset.into(), diagnostics);
         true
     }
+
+    fn register_ir_model_record(&self, session: &mut SessionInfo, record: XmlRecordKey) {
+        let xml_record_sym = &session.st()[record];
+        if xml_record_sym.model.0.as_str() != "ir.model" {
```
cammarosano: no need for `as_str()`, as OYarn implements PartialEq against &str

server/src/core/xml_arch_builder_rng_validation.rs
```diff
@@ -711,4 +715,102 @@ impl XmlArchBuilder {
...
+            return;
+        }
+
+        let maybe_model_name = xml_record_sym
+            .fields()
+            .iter()
+            .find_map(|(name, &field_key)| {
+                if name.as_str() == "model" {
```
cammarosano: same

server/src/core/model.rs
```diff
@@ -134,19 +198,22 @@ impl Model {
...
+            .iter_valid(st)
+            .filter(|&key| {
+                let model = st[key]._model.as_ref().unwrap();
+                !model.inherit.contains(&model.name)
+            })
+            .map(|class_key| class_key.into())
+            .chain(
+                self.xml_model_symbols
+                    .iter_valid(st)
+                    .map(|xml_key| xml_key.into()),
```
cammarosano:
```suggestion
                    .map(SymbolKey::from),
```

server/src/features/goto_utils.rs
```diff
@@ -89,10 +89,11 @@ impl GotoUtils {
         let mut model_found = false;
         let from_module = session.st().find_module(file_symbol);
         let classes = model.borrow().get_symbols(session.st(), from_module);
-        let len_classes = classes.len();
-        for class_key in classes {
+        let xml_model_syms = model.borrow().get_xml_model_symbols(session.st(), from_module);
+        let len_syms = classes.len() + xml_model_syms.len();
+        for class_key in classes.into_iter().map(Into::<SymbolKey>::into).chain(xml_model_syms.into_iter().map(Into::<SymbolKey>::into)) {
```
cammarosano:
```suggestion
        for class_key in classes.into_iter().map(SymbolKey::from).chain(xml_model_syms.into_iter().map(Into::<SymbolKey>::into)) {
```

server/tests/test_get_symbol.rs
```diff
@@ -822,3 +822,254 @@ fn test_csv_ranges_quoted_crlf(session: &mut SessionInfo) {
         "state_quoted_crlf_1 not found in xml_id_locations — CRLF + quoting handling issue"
     );
 }
+/// Test that hover over XML symbols (model name, record id, field name, ref) shows
+/// associated XML record blocks, and that Python hover over model classes and fields
+/// also shows linked XML record blocks when ir.model / ir.model.fields data exist.
+#[test]
+fn test_xml_hover() {
```
cammarosano, line 829: ❤️

server/tests/test_get_symbol.rs
```diff
@@ -822,3 +822,254 @@ fn test_csv_ranges_quoted_crlf(session: &mut SessionInfo) {
...
+    );
+    assert!(
+        hover_py_ref_str.contains("xml_test_model"),
+        "Python hover on xml_id ref string should reference the xml-id; got:\n{}", hover_py_ref_str
+    );
+}
+
+/// Test go-to-definition for XML model name and ref attributes, and from Python model name strings.
+#[test]
+fn test_xml_definition() {
```
cammarosano, line 975: ❤️

server/src/core/xml_validation.rs
```diff
@@ -90,48 +90,72 @@ impl XmlValidator {
...
-        let all_fields = self.fields_cache.get(&model_name).unwrap();
+        let py_fields = self.fields_cache.get(&model_name).into_iter().flat_map(|m| m.keys().cloned());
+        let xml_field_symbols = model.borrow().get_xml_field_symbols(session.st(), Some(self.module));
+        let xml_fields = xml_field_symbols.into_iter().filter_map(|sym| {
+            let record = &session.st()[sym];
+            let Some((_, &field)) = record.fields().into_iter().find(|(field_name, _)| field_name.as_str() == "name") else { return None; };
+            let field = &session.st()[field];
+            field.text.clone().map(|t| Sy!(t))
+        });
+        let all_fields = py_fields.chain(xml_fields).collect::<HashSet<_>>();
```
cammarosano: `py_fields` evaluation depends on `model_name` and `self.module`(which does not change)
`xml_fields` evaluation depends on `model`, which is resolved from `model_name`.
So, both depend only on `model_name` in the end, and their results gets merged to form a new all_fields for each `validate_record` call. Can't you just cache this whole merged result instead of just the py part?

server/src/core/xml_validation.rs
```diff
@@ -61,7 +61,7 @@ impl XmlValidator {
         for data_key in session.st()[self.xml_symbol].symbols().iter().cloned().collect::<Vec<_>>() {
             self.validate_data(session, data_key, &mut diagnostics, &mut dependencies, &mut model_dependencies, &mut missing_model_dependencies);
         }
-        for dep in dependencies {
+        for dep in dependencies.into_iter().filter(|dep_sym| *dep_sym != self.xml_symbol.into()) {
```
cammarosano: 👍

cammarosano (commented):
Very nice! Good job.

.vscode/launch.json
fda-odoo: I cancelled the pr, be sure to remove this line and test that it still work

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -67,7 +68,7 @@ impl SymbolTable {
             SymbolKey::Function(k) => &self[k].name,
             SymbolKey::Variable(k) => &self[k].name,
             SymbolKey::XmlFile(k) => &self[k].name,
-            SymbolKey::XmlRecord(_) => panic!("XmlRecord doesn't have a name"), //TODO do not have a panic?
+            SymbolKey::XmlRecord(r) => self[r].xml_id.as_ref().expect("name() called on XmlRecord without xml_id"),
```
fda-odoo: do you have a specific use case that is using that? Because to me, `name` and `xml_id` are two different things and `xml_id` should not be returned by the `name()` function

server/src/core/model.rs
```diff
@@ -71,6 +73,8 @@ impl ModelData {
 pub struct Model {
     name: OYarn,
     symbols: WeakSet<ClassKey>,
+    xml_model_symbols: WeakSet<XmlRecordKey>,
```
fda-odoo: I don't really like the idea of having two fields (`symbols` and `xml_model_symbols`) that have the exact same meaning, but require a different type. It will imply that everywhere you use `symbols`, you'll have to duplicate the logic for `xml_model_symbols`.
I would rather use super-type (`SymbolKey`, a new one for this case or even `OneOf`) and use `match` later, and handle type locally.
What do you think?

server/src/core/model.rs
```diff
@@ -71,6 +73,8 @@ impl ModelData {
...
+    xml_field_symbols: WeakSet<XmlRecordKey>,
```
fda-odoo, line 78: why do you have to directly add fields on the model? adding records should be enough, no? like we do for python ones?

mmahrouss: Because in XML fields are their own records, while in python they live on the symbols of the class. 
Fields defined in XML are directly linked to the model itself, not to any specific symbol

server/src/core/python_arch_eval_hooks.rs
```diff
@@ -872,14 +873,17 @@ impl PythonArchEvalHooks {
             let model = model.clone();
             let model = model.borrow();
             let symbols = model.get_main_symbols(session, from_module);
-            if let Some(&first_symbol) = symbols.first() {
+            let xml_symbols = model.get_xml_model_symbols(session.st(), from_module);
+            let maybe_first_sym = symbols.first().copied().map(SymbolKey::from).or_else(|| xml_symbols.first().copied().map(SymbolKey::from));
```
fda-odoo: That's the kind of code I would avoid. I guess it's useful if you want python to have precedence on xml models, but I would still iterate over the same list

server/src/core/xml_validation.rs
fda-odoo: How could a file depend on itself?

fda-odoo: moreover this test is already tested by "add_dependency"

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: It was for `build_markdown_description` but then I made a different parsing for XML blocks, so I will revert this part, my bad

server/src/core/xml_validation.rs
mmahrouss: Yes seems it was to avoid a borrow error, but now it does not exist and add dependency checks it, I will remove it. thanks for explaining

server/src/core/xml_validation.rs
```diff
@@ -90,48 +90,72 @@ impl XmlValidator {
...
+        let model_name = model_name.clone();
         if !self.fields_cache.contains_key(&model_name) {
-            let all_fields = SymbolTable::all_fields(main_symbol.into(), session, Some(self.module));
+            let py_fields = main_symbols.get(0).map(|&main_symbol| {
+                SymbolTable::all_fields(main_symbol.into(), session, Some(self.module))
+            });
+            let xml_field_symbols = model.borrow().get_xml_field_symbols(session.st(), Some(self.module));
+            let xml_fields = xml_field_symbols.into_iter().filter_map(|sym| {
+                let record = &session.st()[sym];
+                let Some((_, &field)) = record.fields().into_iter().find(|(field_name, _)| *field_name == "name") else { return None; };
```
fda-odoo: I would call it xml_model_field_symbols to keep the code more clear.
It's a little bit hard to distinguish when we are talking about a `<field/>` and a model field declared in a XmlRecord

fda-odoo:
Is it possible to inherit another model in xml model? You don't seem to handle the concept of "main symbols" from xml models, or am I wrong?

server/src/core/symbols/storage/xml/xml_field_symbol.rs
```diff
@@ -44,4 +44,26 @@ impl XmlFieldSymbol {
     pub fn children(&self) -> Vec<SymbolKey> {
         vec![]
     }
-}
\ No newline at end of file
+}
+
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
```
mmahrouss, line 51: @ Reviewers: What do you think about the location of this struct?

server/src/core/symbols/storage/xml/xml_record_symbol.rs
```diff
@@ -45,4 +47,18 @@ impl XmlRecordSymbol {
...
+
+    pub fn set_declared_model(&mut self, model: OYarn) {
+        self.declared_model = Some(model);
+    }
+
+    pub fn get_declared_model(&self) -> Option<&OYarn> {
+        self.declared_model.as_ref()
+    }
+
+    pub fn get_field_text(&self, field_name: XmlFieldName, symbol_table: &SymbolTable) -> Option<String> {
```
mmahrouss, line 50: @ Reviewers: Do you think this is a good place to place this method? open to suggestions

server/src/core/csv_validation.rs
```diff
@@ -55,15 +53,28 @@ impl CsvValidator {
...
+            let Some(model_main_sym) = model_main_sym.find_map(|s| s.as_class_key()) else {
+                return;
+            };
+            model_main_sym
+        };
         if rdr.has_headers() && let Ok(header) = rdr.headers() {
             let mut header_is_xml = vec![false; header.len()];
             for (idx, (start, end, h)) in CsvFieldIter::new(header, &data).unwrap().enumerate() {
                 let mut header_elts = h.splitn(2, [':', '/']).collect::<Vec<_>>();
                 header_elts[0] = header_elts[0].split("@").next().unwrap(); //remove translation if exists
```
cammarosano: I know this is not quite the subject of your PR, but while you're at it...
This creates a mutable vector, just so that we can mutate its first element and assign its value to `field_name`... I don't see the reason for such complication. If you agree, just evaluate `field_name` based on `header_elts` and the resulting split over `@`.

server/src/core/evaluation_utils.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub fn new(base_object: SymbolKey, from_module: Option<ModuleKey>) -> Self {
+        Self {
+            base_object,
+            from_module,
+            prev_field_symbols: None,
+        }
+    }
+
+    /// Find the next model symbol by looking for relational fields in the current field symbols
+    /// Initial step of the walker
```
cammarosano, line 33: Here you say this method is the initial step of the walker, but on csv_validation you call `get_model_fields` first, then only later you call `get_model_symbol`.

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -66,26 +74,104 @@ impl SymbolTable {
...
+            SymbolKey::XmlAsset(_) => None,
+            SymbolKey::XmlDelete(_) => None,
+        }
+    }
+
+    pub fn name<T>(&self, target: T) -> &OYarn
+    where
+        T: Into<SymbolKey> + Copy,
+    {
+        match self.try_name(target) {
```
cammarosano: considering `try_name` has only this caller, and assuming this call is the only reason you need the `Copy` marker/trait (otherwise it would move a value you need later), I would:
- change the `try_name` signature to take simply `SymbolKey`
- call `target.into()` here (passes the expected SymbolKey to `try_name`)
Then this would allow us to have the simpler `target: impl Into<SymbolKey>` in the signatures or `repr` and `name` (no need for T generic param, no need for + Copy).

server/src/core/csv_validation.rs
```diff
@@ -72,18 +83,9 @@ impl CsvValidator {
...
-                            //on relational fields, the header must contain :id to search for xml_id.
-                            //Else, it will use _rec_name and _rec_name_search to find the right record, which we are not validating
-                            is_relational = true;
-                            break;
-                        }
-                    }
-                    header_is_xml[idx] = is_relational;
+                    header_is_xml[idx] = header_elts.len() == 2
+                        && header_elts[1] == "id"
+                        && deep_field_walker.get_model_symbol(session).is_some();
```
cammarosano: I'm not sure I got this right, but this feels like the only place where the new `DeepFieldEvalWalker` pattern doesn't fit quite right.
You call its methods in the reverse order the doc says, and this call to `get_model_symbol` does more than the original `is_specific_field` does, calling `get_relational_model` and its follow_ref etc.

But it does make the code cleaner, which is good.
So, I'm fine as it is, if you prefer to keep it.

server/src/core/evaluation_utils.rs
```diff
@@ -0,0 +1,134 @@
...
+    },
+    threads::SessionInfo,
+};
+
+/// Helper struct to walk through relational fields of a model in order to find the symbol of a nested field
+/// It follows the chain of relational fields using field string names
+/// It is used by alternating the call between get_model_symbol and get_model_fields,
+/// first to find the model symbol of the next field, then to find the field symbols of the next field.
+/// Gives the possibility to process the fields or the intermediate models at each step.
+pub struct DeepFieldEvalWalker {
```
cammarosano, line 17: This is very cool, I love how much code repetition you removed with this! And also the improvement in semantics/intent/readability.

server/src/features/completion.rs
```diff
@@ -1068,55 +1072,58 @@ fn add_nested_field_names(
...
+            }
+        } else {
+            let field_symbols = deep_field_walker.get_model_fields(session, base_symbol, name);
+            if !add_date_completions {
+                continue;
+            }
+            for symbol in field_symbols {
+                match symbol {
+                    SymbolKey::Variable(_) => {
+                        if SymbolTable::is_specific_field(session, symbol, &["Date", "Datetime"]) {
```
cammarosano, line 1114: ? I suppose this fixes something.

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1496,15 +1582,27 @@ impl SymbolTable {
         only_fields: bool,
         only_methods: bool,
         from_module: Option<ModuleKey>,
+        only_in_deps: bool,
```
cammarosano: if `only_in_deps` is `true` is all call sites and we don't intend to make use of `false` scenario, I would vote for removing this arg.

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1496,15 +1582,27 @@ impl SymbolTable {
...
         is_super: bool
     ) -> HashMap<OYarn, Vec<(SymbolKey, Option<OYarn>)>> {
```
cammarosano: I know it has nothing to do with this PR, but while you're at it... Can you include in the docs what the 2nd member in the tuple represents? I find it rather inintuitive (a missing dependency?).

server/src/features/completion.rs
```diff
@@ -1134,7 +1141,16 @@ fn add_model_attributes(
...
 ){
-    let all_symbols = SymbolTable::all_members(parent_sym, session, true, only_fields, only_methods, from_module, is_super);
+    let all_symbols = SymbolTable::all_members(
+        parent_sym,
+        session,
+        true,
+        only_fields,
+        only_methods,
+        from_module,
+        true,
```
cammarosano: I think you should include a test for this fix (does it fix completion in nested fields? shouldn't be too hard to write a test)

server/src/core/model.rs
```diff
@@ -70,93 +74,152 @@ impl ModelData {
...
+        from_module: Option<ModuleKey>,
+    ) -> impl Iterator<Item = XmlRecordKey> {
+        Self::filter_by_module(
+            self.xml_field_symbols.iter_valid(symbol_table),
+            symbol_table,
+            from_module,
+        )
+    }
+
+    fn get_python_symbols(&self, symbol_table: &SymbolTable) -> impl Iterator<Item = ClassKey> {
```
cammarosano, line 153: very nice that you're returning iterators when possible! S2

cammarosano (approved):
This looks pretty good!
Sorry, but I did not have time to review everything, so this is a first pass. If this has to be merged soon, no worries.
If not merged until next week, I'll review it again in depth.

About commits split for review: really helpful, thanks! But once ready to merge, I think they should be squashed (so that each commit compiles, which is useful for bissecting).

server/src/core/evaluation_utils.rs
mmahrouss: It is because on the first iteration, the base class is what we construct with, and in csv, we only do one level, so it is a no-op in that case, but not when we are working in a loop.

server/src/features/completion.rs
mmahrouss: Yes, before it has Date only, but DateTime is also valid here

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: I would be okay with that, if @fda-odoo agrees

server/src/core/csv_validation.rs
mmahrouss: Here we do not call `get_model_symbol` because it is one iteration, so the first call will only return `model_main_class_sym` which we just passed.

mmahrouss: Previous code was checking: 
1. base field is relational
2. subfield is "id" 
The walker already checks if the field is relational, if its not then the `get_model_symbol` returns None, because there is no more relational trail to follow. 

I check "id" first because it is cheaper. then I test if I have any relational fields. 

What could be different is that the walker only handles the case when we have more than one main symbol from model. which is not handled here. 

Our options are: 
1. add a method somewhere that checks if field is relational, and use that instead.
2. implement the todo for having multiple model symbols
3. let it be for now

What do you think?

server/src/core/csv_validation.rs
mmahrouss: yes will do that thanks

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: yes you are right I found a way to do it, I had to change more things than you mentioned but it works

server/src/core/evaluation_utils.rs
mmahrouss: thank you:)

server/src/features/completion.rs
mmahrouss: the fix is more at `add_nested_field_names`, I will work on a test for that sure

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: As discussed with FDA we can drop it

server/src/core/symbols/storage/xml/xml_record_symbol.rs
fda-odoo: Yes, it's ok for me. Some similar methods are on SymbolTable, but keeping it here make it maybe easier to find?

server/src/core/symbols/storage/xml/xml_field_symbol.rs
cammarosano: Ok here

server/src/core/symbols/storage.rs
```diff
@@ -48,6 +48,9 @@ pub struct SymbolTable {
     xml_deletes: SlotMap<XmlDeleteKey, XmlDeleteSymbol>,
     // external symbols
     ext_symbols: ExtSymbolStore,
+    // secondary slotmaps
+
```
cammarosano: evil empty line 😈

server/src/core/symbols/symbol_keys.rs
```diff
@@ -2,24 +2,26 @@ use slotmap::{Key, new_key_type};
...
+    pub struct VariableKey;
+    pub struct XmlFileKey;
+    pub struct XmlRecordKey;
+    pub struct XmlFieldKey;
+    pub struct XmlMenuItemKey;
+    pub struct XmlTemplateKey;
+    pub struct XmlAssetKey;
+    pub struct XmlDeleteKey;
+    pub struct CsvFileKey;
+}
```
cammarosano, line 24: amazing!

server/src/core/symbols/storage/xml/xml_record_symbol.rs
```diff
@@ -45,4 +45,21 @@ impl XmlRecordSymbol {
...
+}
+
+impl SymbolTable {
+    pub fn set_declared_model(&mut self, xml_record: XmlRecordKey, model: OYarn) {
+        self.xml_declared_models.insert(xml_record, model);
+    }
+
+    pub fn get_declared_model(&self, xml_record: XmlRecordKey) -> Option<&OYarn> {
+        self.xml_declared_models.get(xml_record)
+    }
```
cammarosano, line 64: Every call to this seems to be followed by `session.sync_odoo.models.get(model_name)`.
Consider adding a function (`get_xml_defined_model` or something, maybe in symbol_table_impl?) that takes `&SyncOdoo`, does both steps and returns `Option<Rc<RefCell<Model>>>`. I think it would simplify the call sites.

server/src/features/xml_ast_utils.rs
```diff
@@ -99,11 +98,20 @@ impl XmlAstUtils {
...
+                        for class_key in Model::get_full_model_classes(model.clone(), session, from_module) {
+                            let content = session.st().get_content_symbol(class_key.into(), field_name, u32::MAX);
+                            for symbol in content.symbols {
+                                results.0.push(symbol);
+                            }
+                        }
+                        let model_ref = model.borrow();
+                        for xml_record_key in model_ref.get_xml_model_field_symbols(session.st(), from_module) {
+                            let record = &session.st()[xml_record_key];
+                            if let Some(&name_field_key) = record.fields().get(&Sy!("name")) {
```
cammarosano: You can do lookups in HashMaps keyed by OYarn with a `&'static str`, no need to build a OYarn here: 
```suggestion
                            if let Some(&name_field_key) = record.fields().get("name")) {
```

cammarosano (approved):
Very good job! Thanks for the patience and explanations.

server/src/core/csv_validation.rs
cammarosano: I prefer option 1, I think it would make the code more readable/declarative.
But also fine if you would rather keep it like this.

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1572,16 +1659,36 @@ impl SymbolTable {
                     //TODO what about base of co-models classes?
                     Self::_all_members(base.into(), session, result, false, only_fields, only_methods, from_module, acc, false);
                 }
-            },
+            }
+            SymbolKey::XmlRecord(xml_record_key) => {
+                // If it is a field-defining record
```
cammarosano: did you mean "model-defining record"?

mmahrouss: Yes!

server/src/core/symbols/storage/xml/xml_record_symbol.rs
mmahrouss: agreed

server/src/core/csv_validation.rs
mmahrouss: Done:)

server/src/core/model.rs
```diff
@@ -70,93 +74,152 @@ impl ModelData {
...
+            .filter_map(|sym| sym.as_class_key())
+    }
+
+    fn filter_by_module<T: Into<SymbolKey> + Copy>(
+        iter: impl Iterator<Item = T>,
+        symbol_table: &SymbolTable,
+        from_module: Option<ModuleKey>,
+    ) -> impl Iterator<Item = T> {
+        iter.filter(move |sym| {
+            let module = symbol_table
```
fda-odoo: `let module = ` could be in `Some(module_key)` case

mmahrouss: indeed, but actually I think we should actually  pass the modulekey directly not an option, and only call this filter if we have a module. but I agree here we can get rid of the expect

fda-odoo (approved):

server/src/core/model.rs
mmahrouss: Okay as explained irl, I misunderstood the comment, I will move it to the match statement yes!

PR #603, opened 2026-05-18
Skip unavailable CSV/XML files, and only track the file changes of relevant files

mmahrouss:


server/src/core/xml_validation.rs
```diff
@@ -27,11 +27,20 @@ impl XmlValidator {
         }
     }
 
-    fn get_file_info(&mut self, odoo: &mut SyncOdoo) -> Rc<RefCell<FileInfo>> {
+    fn get_file_info(&mut self, session: &mut SessionInfo) -> Option<Rc<RefCell<FileInfo>>> {
```
cammarosano: consider making this a shared function that takes the `Rc<RefCell<Symbol>>` as arg, if you find it suitable.

server/src/core/csv_validation.rs
```diff
@@ -17,10 +18,30 @@ impl CsvValidator {
...
     }
 
+
+    fn get_file_info(session: &mut SessionInfo, csv_symbol: Rc<RefCell<Symbol>>) -> Option<Rc<RefCell<FileInfo>>> {
+        let file_symbol = csv_symbol.borrow();
+        let path = file_symbol.paths()[0].clone();
+        match session.sync_odoo.get_file_mgr().borrow().get_file_info(&path).clone() {
+            Some(file_info) => Some(file_info),
+            None => {
+                let (updated, symbol) = session.sync_odoo.get_file_mgr().borrow_mut().update_file_info(session, &file_symbol.paths()[0], None, Some(-100), true);
```
cammarosano: you could just use the already computed (and cloned) `path` here, no?


```suggestion
                let (updated, symbol) = session.sync_odoo.get_file_mgr().borrow_mut().update_file_info(session, &path, None, Some(-100), true);
```

And, maybe `path` could be simply `&file_symbol.paths()[0]` (no clone)?

cammarosano (approved):
Nice!
Some minor nit 😬

server/src/core/xml_validation.rs
cammarosano: (same comment about using `path` applies here, if you don't merge the 2 functions into a shared one).

mmahrouss: Yes you are right, I combined them with the one in python validator into a shared fn.

cammarosano (approved):

fda-odoo (approved):

PR #607, opened 2026-05-18
[FIX] dedup bugs in references engine and model walker

andg-odoo:
Two small walker bugs in the language server, both surfacing as duplicated or missed results in references / goto-def / hover.

- **References dedup by start position** (6f88cd5): the visitor deduped each new hit against the previous entry *by line*, collapsing multiple same-line uses (e.g. `if x is None or 'a' in x or 'b' in x` kept only the first `x`). The line-based dedup also masked a separate issue: nested expressions record at the same start position from different recursion depths. Switch to `(uri, range.start)` over the full list. Innermost is analyzed first, so the most specific token wins. `test_references` gains one expected entry (find-refs on the model now records `self` in `self.env[...]`).

- **Stop `Model::all_symbols` from re-walking self** (5367c5e): `all_symbols_helper` tracked visited *inherited* models but never marked the current model. Inheriting classes carry the current model in their own `_inherit`, so the recursion re-entered self once per inheriting class, doubling every entry. Symptoms: two LocationLinks from goto-def on `<field name="X"/>`, doc string rendered twice on hover, validation walking every class N+1 times. Mirror the self-visited check `all_inherits_helper` already had.

### DIFF
--- a/server/src/core/evaluation.rs
+++ b/server/src/core/evaluation.rs
@@ -256,6 +256,23 @@ impl AnalyzeAstResult {
     }
 }
 
+/// Record a hit for the active `evaluation_search` at `range` inside `parent`'s file.
+/// Innermost expressions are analyzed first, so the first record at a given start
+/// position wins; outer enclosing expressions starting at the same place are skipped.
+/// Different start positions on the same line are kept (for multiple uses of a name).
+fn record_evaluation_hit(session: &mut SessionInfo, parent: SymbolKey, range: TextRange) -> bool {
+    let Some(file) = session.st().get_file(parent) else { return false };
+    let file_path = session.st().path(file).to_string();
+    let Some(file_info) = session.sync_odoo.get_file_mgr().borrow().get_file_info(&file_path) else { return false };
+    let transformed_range = file_info.borrow().text_range_to_range(&range, session.sync_odoo.encoding);
+    let uri = FileMgr::pathname2uri(&file_path);
+    if session.sync_odoo.evaluation_locations.iter().any(|loc| loc.uri == uri && loc.range.start == transformed_range.start) {
+        return false;
+    }
+    session.sync_odoo.evaluation_locations.push(Location { uri, range: transformed_range });
+    true
+}
+
 impl Evaluation {
 
     pub fn new_list(odoo: &mut SyncOdoo, values: Option<Vec<Expr>>, range: TextRange) -> Evaluation {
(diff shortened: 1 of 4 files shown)


server/tests/data/addons/module_1/models/base_test_models.py
```diff
@@ -61,4 +61,9 @@ class NoBaseModel(models.Model):
 dict_ref = {1: basic_var}
 call_ref = print(basic_var)
 binop_ref = basic_var + 1
-lambda_scope = lambda basic_var: basic_var
\ No newline at end of file
+lambda_scope = lambda basic_var: basic_var
+
+def annotated_param_func(field_names: list | None = None) -> None:
```
fda-odoo, line 66: what is the purpose of this function in tests?

server/src/core/evaluation.rs
```diff
@@ -1521,32 +1538,17 @@ impl Evaluation {
...
-                            if session.sync_odoo.evaluation_locations.is_empty() ||
-                            session.sync_odoo.evaluation_locations.last().unwrap().uri != uri ||
-                            session.sync_odoo.evaluation_locations.last().unwrap().range.start.line != transformed_range.start.line {
-                                session.sync_odoo.evaluation_locations.push(Location {
-                                    uri: uri,
-                                    range: transformed_range,
-                                });
-                            }
-                        }
+                        found_one_reference = true;
```
fda-odoo: found_one_reference was not set to true if file_info is not found.
Maybe you could use the return value of record_evaluation_hit to set it?
```rust
found_one_reference |= record_evaluation_hit(session, parent, ast.range());
```

server/src/core/evaluation.rs
```diff
@@ -1555,41 +1557,15 @@ impl Evaluation {
...
-                                            if session.sync_odoo.evaluation_locations.is_empty() ||
-                                            session.sync_odoo.evaluation_locations.last().unwrap().uri != uri ||
-                                            session.sync_odoo.evaluation_locations.last().unwrap().range.start.line != transformed_range.start.line{
-                                                session.sync_odoo.evaluation_locations.push(Location {
-                                                    uri: uri,
-                                                    range: transformed_range,
-                                                });
-                                            }
-                                        }
+                                        found_one_reference = true;
```
fda-odoo: same here

fda-odoo (changes requested):
Hey @andg-odoo ,
That's a pleasure to see these pull requests ! This one seems good to me, except the small comments.
Thank you for submitting it !

server/tests/data/addons/module_1/models/base_test_models.py
andg-odoo: I forgot to push the hunk that tested this function. it is basically just checking that we can find all references on the same line. Most of the time it is not necessary currently but if a rename feature is added later then we need to find all occurrences on a line not only the first. Tests have been added.

andg-odoo:
@fda-odoo Fixed the find_one_reference and re-introduced the tests that were forgotten.

andg-odoo:
:wave: @fda-odoo let me know if there is anything else necessary to change for this pr :)

fda-odoo:
@andg-odoo It looks good to me! Thank you for the pull request !

fda-odoo (approved):

PR #608, opened 2026-05-18
[IMP] XML resolution: arch fields, buttons, format-string refs

andg-odoo:
Walker corrections and new XML constructs handled end-to-end by goto-def / hover / find-refs. Final commit consolidates helpers introduced earlier.

Rebased onto current `alpha-1.5.0` (post-#600 / symbol-table refactor): `resolve_member_on_model` now uses `Model::get_full_model_classes` + `get_xml_model_field_symbols`, so fields defined in XML (`ir.model.fields` records) resolve too.

- **`[IMP] recognize %(xml_id)d format-string references in XML`**: `<button name="%(module.action_x)d" type="action"/>` and similar `%(...)s|i` patterns now resolve via goto-def / hover and match in find-refs. The reported range covers exactly the inner xml-id (excluding the `%(` / `)d` wrapper), so a future rename feature can rewrite just the id.

- **`[FIX] resolve field/method names inside arch field against the views target model`**: the walkers kept using the outer record's `model=` for the whole subtree, so goto-def on `<field name="email"/>` inside an `ir.ui.view` searched `email` on `ir.ui.view` itself. When entering an arch field, sibling `<field name="model">X</field>` swaps the active `record_model` to X for the subtree, save/restored for nesting.

- **`[IMP] resolve <button name=method_x> against the surrounding record model`**: buttons with implicit or explicit `type="object"` invoke a Python method on the surrounding record's model. Previously goto-def / hover did nothing and find-refs missed every callsite. Add the `<button>` case to both walkers, resolving `name` against `record_model` when type is not `"action"`. Also extends the references engine to walk model dependents for FUNCTION targets (previously only CLASS and field-Variable did).

- **`[FIX] register inner field model as XML dependency for view-like records`**: view-like records (`ir.ui.view`, `ir.actions.act_window`, etc.) were only dependents of their outer container model. find-refs from a method on the *target* model therefore never walked the view XML, so `<button>` / `<field>` callsites in the view were invisible to references. Validation now pushes the inner `<field name="model">` value into `model_dependencies` whenever it resolves. Covers all view-like containers without an allow-list.

- **`[REF] consolidate shared XML helpers and dedup references`**: the format-string scanner, arch-field target-model lookup, and button-resolves-to-method predicate had byte-identical copies in `references_xml.rs` and `xml_ast_utils.rs`. Moved to a new `features/xml_common.rs`. Also extracts `resolve_member_on_model` shared between the field and button walkers, routes XML-field name/model lookups through the typed `XmlRecordSymbol::get_field_text` accessor, and replaces a per-push O(N) dedup in `references_in_file` with sort+`dedup_by` at end of walk.

### DIFF
--- a/server/src/features/references_xml.rs
+++ b/server/src/features/references_xml.rs
@@ -52,6 +52,9 @@ impl XmlAstReferenceVisitor {
                 "template" => {
                     XmlAstReferenceVisitor::visit_template(session, &node, from_module, ctxt, results, target);
                 }
+                "button" => {
+                    XmlAstReferenceVisitor::visit_button(session, &node, from_module, ctxt, results, target);
+                }
                 _ => {
                     for child in node.children() {
                         XmlAstReferenceVisitor::visit_node(session, &child, from_module, ctxt, results, target);
@@ -60,6 +63,33 @@ impl XmlAstReferenceVisitor {
             }
         } else if node.is_text() {
             XmlAstReferenceVisitor::visit_text(session, &node, from_module, ctxt, results, target);
+        }
+    }
+
+    /// `<button name="method_x" type="object">` invokes a method on the current
+    /// record's model. When the rename/find-refs target is a method (Function on
+    /// a model class), match the button's `name` value against the method name
+    /// AND verify the surrounding `record_model` matches the method's class model.
+    fn visit_button(session: &mut SessionInfo<'_>, node: &Node, from_module: Option<ModuleKey>, ctxt: &mut HashMap<String, ContextValue>, results: &mut Vec<Range<usize>>, target: &ReferenceTarget) {
+        let is_action_type = node.attribute("type") == Some("action");
+        if !is_action_type {
+            if let &ReferenceTarget::Symbol(SymbolKey::Function(target_fn)) = target {
+                for attr in node.attributes() {
+                    if attr.name() != "name" { continue; }
+                    if session.st()[target_fn].name != attr.value() { continue; }
+                    let target_class = session.st().get_in_parents(target_fn.into(), &[SymType::CLASS], true);
+                    let Some(SymbolKey::Class(target_class)) = target_class else { continue; };
+                    let Some(target_model) = session.st()[target_class]._model.as_ref() else { continue; };
+                    let record_model = ctxt.get("record_model").map(ContextValue::as_str).unwrap_or_default();
+                    if record_model.is_empty() { continue; }
+                    if target_model.name == *record_model {
+                        results.push(attr.range_value());
+                    }
+                }
+            }
+        }
+        for child in node.children() {
+            XmlAstReferenceVisitor::visit_node(session, &child, from_module, ctxt, results, target);
         }
     }
 
--- a/server/src/features/references_xml.rs
+++ b/server/src/features/references_xml.rs
@@ -5,7 +5,7 @@ use crate::{
         symbols::{
             storage::SymbolTable, symbol_keys::{ModuleKey, SymbolKey, XmlFileKey}
         },
-    }, features::references::ReferenceTarget, threads::SessionInfo
+    }, features::{references::ReferenceTarget, xml_ast_utils::XmlAstUtils}, threads::SessionInfo
 };
 use lsp_types::Location;
 use roxmltree::Node;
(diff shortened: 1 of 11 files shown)


server/src/features/xml_ast_utils.rs
```diff
@@ -117,12 +117,46 @@ impl XmlAstUtils {
...
+                Some(prev) => { ctxt.insert(S!("record_model"), prev); }
+                None => { ctxt.remove("record_model"); }
+            }
+        }
         ctxt.remove("field_name");
     }
 
+    /// If `node` is `<field name="arch">`, return the text of its sibling
+    /// `<field name="model">…</field>` (the view's target model).
+    fn pick_arch_target_model(node: &Node) -> Option<String> {
```
fda-odoo: I don't really like the way your are searching for arch tags for ir.ui.view, even if I guess it's working most of the time
1. We usually try to avoid using contexts, that are hashing strings and make the code more 'spaghetti'. I'm pretty sure we can avoid it here.
2. you are trying to search for an "arch" field, regardless of the record_model, and so you don't test if we are in an ir.ui.view. This code will be applied on any arch on any model.

In xml files, I would go from top to bottom only: detect ir.ui.view on a node, so search for model, then for arch and evaluate it as such

server/src/features/references_xml.rs
```diff
@@ -66,25 +66,20 @@ impl XmlAstReferenceVisitor {
...
-                    if session.st()[target_fn].name != attr.value() { continue; }
-                    let target_class = session.st().get_in_parents(target_fn.into(), &[SymType::CLASS], true);
-                    let Some(SymbolKey::Class(target_class)) = target_class else { continue; };
-                    let Some(target_model) = session.st()[target_class]._model.as_ref() else { continue; };
-                    let record_model = ctxt.get("record_model").map(ContextValue::as_str).unwrap_or_default();
-                    if record_model.is_empty() { continue; }
-                    if target_model.name == *record_model {
-                        results.push(attr.range_value());
-                    }
+        'resolve: {
```
fda-odoo: I don't really like this refactoring. I feel that it makes the code messy, with all these 'resolve

server/src/features/references_xml.rs
```diff
@@ -66,25 +66,20 @@ impl XmlAstReferenceVisitor {
...
+            if !xml_common::button_invokes_method(node) { break 'resolve; }
```
fda-odoo: I don't see why you need to put this condition test in a common method. It means that to see the test I have to follow and read the function, that is only hiding a single test.
Moreover, Docstring of this method is not really explaining the contained test.
I would keep `node.attribute("type") != Some("action")` here

server/src/features/xml_ast_utils.rs
```diff
@@ -62,34 +64,22 @@ impl XmlAstUtils {
...
     }
 
-    /// `<button name="method_x" type="object">` invokes `method_x` on the current
-    /// record's model. Implicit `type` is `object` in views, so we resolve `name`
-    /// against `record_model` whenever there's no `type="action"` (which would put
-    /// us in the xml-id case already handled by scan_format_xml_id_under_cursor /
-    /// the existing %(...)d path).
     fn visit_button(session: &mut SessionInfo<'_>, node: &Node, offset: usize, from_module: Option<ModuleKey>, ctxt: &mut HashMap<String, ContextValue>, results: &mut (Vec<SymbolKey>, Option<Range<usize>>), on_dep_only: bool) {
-        let is_action_type = node.attribute("type") == Some("action");
+        let resolves_to_method = xml_common::button_invokes_method(node);
```
fda-odoo: So you extracted methods from `xml_ast_utils` to `xml_common`. Common of what? it looks like a "utils" of "utils"
xml_ast_utils is already a toolbox, you should keep these functions in there

fda-odoo (changes requested):
4 first commits are nices !  Thank you for the addition.
However, your 5th cleanup commit is not really cleaning things (It seems Ai want to clean what is clean whatsoever). Some deduplication are nice and should be kept (the big ones), but I don't really see the point of deduplicating a single expression like `type == 'action'`
We fixed the tests on alpha-1.5.0 too, so you can rebase on it

server/src/features/xml_ast_utils.rs
andg-odoo: Both points have been addressed in a follow-up commit _([REF] thread XML resolution scope top-down instead of a string context)_

1. The string context is no longer being used and instead has a typed `Copy` `XmlScope` datatype. This simplifies the arch-field save/restore code a lot and string hashing no longer occurs. By doing this a visitor can pass a tweaked copy of the data to it's children that way it's scoped automatically.

2. I've now resolved the arch swap top-down, as you suggested. When the surrounding record is of type `ir.ui.view` its target model is captured at the record and applied to the arch subtree, instead of firing on every `arch` field on any model.

This was kept as a separate commit instead of folding into some of the earlier ones as it requires a resignature on the other visitors.

This seems like the most structurally safe and cleanest way to implement such a fix but if you have other ideas I'm all ears.

fda-odoo:
Ok that's great, we'll take it on our side from here :)

fda-odoo (approved):

PR #609, opened 2026-05-18
[IMP] enable XML hover for non-Class and XML data symbols

andg-odoo:
`hover_xml` filtered the resolved cursor symbols down to `SymbolKey::Class` before rendering, leaving every other XML cursor target (field declarations, xml-id refs, method names) with empty hover. The Python hover already passes everything through `build_markdown_description` for Variables and Functions; dropping the Class-only filter gets those for free.

XML data symbols (`XmlRecord`, `XmlField`, `XmlMenuItem`, `XmlTemplate`, `XmlAsset`, `XmlDelete`) have no `Symbol.name()` and don't fit the existing aggregator, so they get a dedicated rendering path showing the kind, qualified xml-id, declaring module, and model where applicable.

Hovering on `<record id=...>`, `<field name=...>`, `<menuitem id=...>` etc. now produces something useful instead of empty content.

### DIFF
--- a/server/src/features/features_utils.rs
+++ b/server/src/features/features_utils.rs
@@ -528,6 +528,11 @@ impl FeaturesUtils {
                 }
                 continue;
             };
+            // XML data symbols have no Symbol.name() and don't fit the aggregator below.
+            if let Some(block) = FeaturesUtils::xml_data_hover_block(session, symbol) {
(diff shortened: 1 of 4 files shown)


andg-odoo:
Closing in favor of #600, which does this better.

PR #611, opened 2026-05-21
feat: implement Sphinx formatting

Av1d1ty:
**This is made for demonstration purposes only and should NOT be merged. The code is mostly generated.**
This is a follow-up to this issue [issue](https://github.com/Whenrow/odoo-ls.nvim/issues/9).

I wanted odoo-ls to be able to parse Sphinx-style docstrings the same way [python-lsp-server](https://github.com/python-lsp/python-lsp-server) does instead of treating them as plain markdown. I am using this fork for a few weeks now and it seems to work fairly well.

Here is an example of a formatted docstring in a hover window:
image

Feel free to close the PR and write something more thoughtful if you see any value in this feature. But I really hope you would consider implementing sphynx parsing, because most of base odoo docstrings are written in this format.

### DIFF
--- a/server/src/features/features_utils.rs
+++ b/server/src/features/features_utils.rs
@@ -1,4 +1,5 @@
 use itertools::Itertools;
+use regex::Regex;
 use ruff_python_ast::{Expr, ExprCall, Keyword};
 use ruff_text_size::{Ranged, TextRange, TextSize};
 use crate::core::file_mgr::FileMgr;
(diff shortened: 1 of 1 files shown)


fda-odoo:
Hello @Av1d1ty , thank you for your pull request.
We will check into the sphinx formatting and your pull request after the release of 1.5.0, so I can't answer you right now what we will do about it, but we saw your issue and I wanted to let you know that we'll check at it soon :)

PR #612, opened 2026-05-27
[IMP] manifest loading in steps

fda-odoo:
To lower the rebuilds needed when a change is made in data files, the module initialization has been split into:
- symbol creation: we load the manifest content (needed to sort module loading)
- Arch: normal arch python on `__init__.py`
- Arch_eval: load data files (was previously in symbol creation). Load them only if symbol does not already exists.
- Validation: python and manifest validation.


server/src/core/symbols/manifest_create.rs
```diff
@@ -0,0 +1,315 @@
...
+impl ModuleSymbol {
+
+    pub fn load_manifest_content(session: &mut SessionInfo, module_key: ModuleKey) {
+        let manifest_path = PathBuf::from(&session.st()[module_key].path).join("__manifest__.py");
+        if DEBUG_STEPS {
+            info!("ARCH       - MANIFEST: {}", manifest_path.sanitize());
+        }
+        let (_, manifest_file_info) = session.sync_odoo.get_file_mgr().borrow_mut().update_file_info(session, manifest_path.sanitize().as_str(), None, None, false);
+        let mut manifest_file_info = (*manifest_file_info).borrow_mut();
+        //TODO remove me?
```
cammarosano: Shouldn't the call to `update_file_info` call above include `true` for the `force` param, ir order to guarantee that the ast will be prepared?
If I understood this correctly, the AST won't be parsed if the file is opened in the user's editor.

Not parsing the manifest will leave the module's `depends` as an empty vec, which will mess up module load order sort. Have you tested if your branch loads modules in the same order as the base branch? Having a manifest opened??

server/src/core/symbols/manifest_arch.rs
```diff
@@ -0,0 +1,315 @@
...
+        //     manifest_file_info.prepare_ast(session);
+        // }
+        if manifest_file_info.file_info_ast.borrow().indexed_module.is_none() {
+            return;
+        }
+        let diags = ModuleSymbol::load_manifest(session, module_key, &manifest_file_info);
+        if session.sync_odoo.modules.contains_key(&session.st()[module_key].dir_name) {
+            //TODO: handle multiple modules with the same name
+        }
+        manifest_file_info.replace_diagnostics(crate::constants::BuildSteps::SYNTAX, diags);
```
cammarosano: import `crate::constants::BuildSteps` then
```suggestion
        manifest_file_info.replace_diagnostics(BuildSteps::SYNTAX, diags);
```

server/src/core/symbols/manifest_validation.rs
```diff
@@ -0,0 +1,49 @@
...
+                    });
+                }
+                session.st_mut()[module_key].not_found_models.insert(model_name.clone(), BuildSteps::VALIDATION);
+                session.sync_odoo.get_main_entry().borrow_mut().not_found_symbols_for_models.insert(module_key.into());
+            }
+        }
+        let manifest_path = PathBuf::from(root_path).join("__manifest__.py");
+        let manifest_file_info = session.sync_odoo.get_file_mgr().borrow().get_file_info(&manifest_path.sanitize()).expect("file not found in cache").clone();
+        let mut manifest_file_info = (*manifest_file_info).borrow_mut();
+        manifest_file_info.replace_diagnostics(crate::constants::BuildSteps::VALIDATION, diagnostics);
```
cammarosano: `BuildSteps::VALIDATION` is enough here (already imported)

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+                let document = roxmltree::Document::parse(&data);
+                if let Ok(document) = document {
+                    file_info.replace_diagnostics(BuildSteps::SYNTAX, vec![]);
+                    let root = document.root_element();
+                    let mut xml_builder = XmlArchBuilder::new(xml_sym);
+                    xml_builder.load_arch(session, &mut file_info, &root);
+                } else if data.len() > 0 {
+                    let mut diagnostics = vec![];
+                    XmlFileSymbol::build_syntax_diagnostics(&session, &mut diagnostics, &mut file_info, &document.unwrap_err());
+                    file_info.replace_diagnostics(BuildSteps::SYNTAX, diagnostics);
```
cammarosano: shouldn't this be ARCH_EVAL?
As it is, this is going to replace the diagnostics added by `load_manifest_content`, right?
Did you test if your branch emits the same diagnostics as the base branch?

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+            let path = PathBuf::from(module_path.clone()).join(data_url);
+            let file_name = path.file_name().unwrap().to_str().unwrap().to_string();
+            //check if already exists
+            if session.st()[symbol_key].data_symbols().contains_key(&path.sanitize()) {
+                continue;
+            }
+            //load data from file
+            let path = PathBuf::from(module_path.clone()).join(data_url);
+            if !path.exists() {
+                session.st_mut()[symbol_key].not_found_data.insert(path.sanitize(), BuildSteps::ARCH);
```
cammarosano: shouldn't this be ARCH_EVAL?

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+                    let mut diagnostics = vec![];
+                    XmlFileSymbol::build_syntax_diagnostics(&session, &mut diagnostics, &mut file_info, &document.unwrap_err());
+                    file_info.replace_diagnostics(BuildSteps::SYNTAX, diagnostics);
+                    file_info.publish_diagnostics(session);
+                    continue
+                }
+            } else if file_name.ends_with(".csv") {
+                let csv_sym = session.st_mut().add_new_csv_file(symbol_key, &file_name, &path.sanitize());
+                Self::on_data_file_load(session.st(), csv_sym.into());
+                session.st_mut().add_dependency(symbol_key.into(), csv_sym.into(), BuildSteps::ARCH, BuildSteps::ARCH);
```
cammarosano: shouldn't `step` arg be `ARCH_EVAL`, like it is above for xml file?

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn load_data(symbol_key: ModuleKey, session: &mut SessionInfo) {
+        let mut diagnostics = vec![];
+        let module = &session.st()[symbol_key];
+        let module_path = module.path.clone();
+        if DEBUG_STEPS {
+            info!("ARCH_EVAL  - MANIFEST: {}", module_path);
+        }
+        let data_paths = module.data.clone();
+        for (data_url, data_range) in data_paths.iter() {
+            let path = PathBuf::from(module_path.clone()).join(data_url);
```
cammarosano:
```suggestion
            let path = Path::new(&module_path).join(data_url);
```
This avoids the 2 allocations: the PathBuf and the String clone.

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+            let file_name = path.file_name().unwrap().to_str().unwrap().to_string();
+            //check if already exists
+            if session.st()[symbol_key].data_symbols().contains_key(&path.sanitize()) {
+                continue;
+            }
+            //load data from file
+            let path = PathBuf::from(module_path.clone()).join(data_url);
```
cammarosano: This line duplicates what has been done a few lines above. Just use the same `path`.

cammarosano (changes requested):
Nice! But I have a few concerns... or maybe I'm just confused!

server/src/core/symbols/manifest_create.rs
fda-odoo: opened files are always handled after the first loading, so it should not happen. No DidOpen notification can arrive before the db init

fda-odoo:
@cammarosano thank you for the review, I updated the pr with your comments. I tested against 1.5.0, and I have exactly the same diagnostics

server/src/core/symbols/manifest_create.rs
```diff
@@ -0,0 +1,315 @@
...
+        // if manifest_file_info.file_info_ast.borrow().indexed_module.is_none() {
+        //     manifest_file_info.prepare_ast(session);
+        // }
```
mmahrouss: What happens if we create a symbol from an import? is it possible that we need to run prepare ast on it?

mmahrouss: I guess since we run load_depends it will always be already loaded?

server/src/core/symbols/manifest_create.rs
mmahrouss, line 1: Isn't the name `manifest_arch` a bit confusing? since it does not run in the arch step? it runs actually upon creation?

server/src/core/symbols/module_utils.rs
mmahrouss, line 1: Relating to the other naming comment, shouldn't this be the `module_arch` file?

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+            if file_name.ends_with(".xml") {
+                let xml_sym = session.st_mut().add_new_xml_file(symbol_key, &file_name, &path.sanitize());
+                Self::on_data_file_load(session.st(), xml_sym.into());
+                session.st_mut().add_dependency(symbol_key.into(), xml_sym.into(), BuildSteps::ARCH_EVAL, BuildSteps::ARCH);
+                if file_info.file_info_ast.borrow().text_document.as_ref().is_none() {
+                    //TODO do we want to add a diagnostic here?
+                    continue;
+                }
+                //That's a little bit crappy, but the SYNTAX step of XML files are done here, as lifetime of roXMLTree are not flexible enough to be separated from the Arch building
+                let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
```
mmahrouss: Double access of text_document, we can change it to a match block like 

```suggestion
                let data = match file_info.file_info_ast.borrow().text_document.as_ref() {
                    Some(text_document) => text_document.contents().to_string(),
                    None => {
                        //TODO do we want to add a diagnostic here?
                        continue;
                    }
                };
```

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+                Self::on_data_file_load(session.st(), csv_sym.into());
+                session.st_mut().add_dependency(symbol_key.into(), csv_sym.into(), BuildSteps::ARCH_EVAL, BuildSteps::ARCH);
+                if file_info.file_info_ast.borrow().text_document.as_ref().is_none() {
+                    //TODO do we want to add a diagnostic here?
+                    continue;
+                }
+                let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
+                let mut csv_builder = CsvArchBuilder::new();
+                let diagnostics = csv_builder.load_csv(session, csv_sym, &data);
+                file_info.replace_diagnostics(BuildSteps::SYNTAX, diagnostics);
```
mmahrouss, line 84: Do you mean ARCH_EVAL here?

server/src/core/symbols/manifest_create.rs
fda-odoo: it is indeed, it was planned to do it at arch with this refactoring, but I had to change it back to the creation step to keep the module loading order. I didn't rename this file though, I'll do it

server/src/core/symbols/manifest_arch_eval.rs
mmahrouss: Same in  multiple instances in the same file

server/src/core/symbols/manifest_create.rs
fda-odoo: I don't really know why this code was there, as update_file_info will prepare_ast, except if the file is already loaded. But in that case prepare_ast would not be useful...
But the code is there, so there was reason, and I'm afraid to miss something like a cache invalidation somewhere

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+        let data_paths = module.data.clone();
+        for (data_url, data_range) in data_paths.iter() {
+            let path = Path::new(&module_path).join(data_url);
+            let file_name = path.file_name().unwrap().to_str().unwrap().to_string();
+            //check if already exists
+            if session.st()[symbol_key].data_symbols().contains_key(&path.sanitize()) {
+                continue;
+            }
+            //load data from file
+            let path = Path::new(&module_path).join(data_url);
```
mmahrouss: `path` is duplicated
```suggestion
            let path = Path::new(&module_path).join(data_url);
            let file_name = path.file_name().unwrap().to_str().unwrap().to_string();
            //check if already exists
            if session.st()[symbol_key].data_symbols().contains_key(&path.sanitize()) {
                continue;
            }
            //load data from file
```

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+                //That's a little bit crappy, but the SYNTAX step of XML files are done here, as lifetime of roXMLTree are not flexible enough to be separated from the Arch building
+                let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
+                let document = roxmltree::Document::parse(&data);
+                if let Ok(document) = document {
+                    file_info.replace_diagnostics(BuildSteps::SYNTAX, vec![]);
+                    let root = document.root_element();
+                    let mut xml_builder = XmlArchBuilder::new(xml_sym);
+                    xml_builder.load_arch(session, &mut file_info, &root);
+                } else if data.len() > 0 {
```
mmahrouss:
```suggestion
                } else if !data.is_empty() {
```

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+                let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
+                let mut csv_builder = CsvArchBuilder::new();
+                let diagnostics = csv_builder.load_csv(session, csv_sym, &data);
+                file_info.replace_diagnostics(BuildSteps::SYNTAX, diagnostics);
+                file_info.publish_diagnostics(session);
+            }
+        }
+        let manifest_path = PathBuf::from(&module_path).join("__manifest__.py");
+        let Some(manifest_file_info) = session.sync_odoo.get_file_mgr().borrow_mut().get_file_info(&manifest_path.sanitize()) else {
```
mmahrouss:
```suggestion
        let Some(manifest_file_info) = session.sync_odoo.get_file_mgr().borrow().get_file_info(&manifest_path.sanitize()) else {
```

mmahrouss (commented):
Good work:) minor comments

server/src/core/symbols/module_utils.rs
fda-odoo: I would keep "manifest", as these files contain code only for manifests, not the whole module. Steps of `__init__.py` is taking the regular route for example

server/src/core/symbols/manifest_arch_eval.rs
fda-odoo: No, it's the diagnostics from the csv_builder, so it's the SYNTAX level

cammarosano (approved):

PR #615, opened 2026-06-02
[FIX] Fix  borrow error from build_now_dependencies

mmahrouss:
sym reference was held during this function, because of the call to get_all_dependencies, this commit makes sure to collect the needed dependencies first before calling build_now, which avoid the panic
also a test was added to make sure no panic occurs in that scenario

It does not cause any error on alpha-1.5.0, so it is even safe to drop by then, for when rebasing.

see crash report 2155

### DIFF
--- a/server/tests/test_borrow_panic.rs
+++ b/server/tests/test_borrow_panic.rs
@@ -1,0 +1,117 @@
+use std::env;
+use std::path::PathBuf;
+use std::rc::Rc;
+
+use odoo_ls_server::constants::{BuildStatus, BuildSteps, OYarn};
+use odoo_ls_server::core::odoo::SyncOdoo;
+use odoo_ls_server::Sy;
+use odoo_ls_server::utils::PathSanitizer;
+
+mod setup;
+use setup::setup::{create_init_session, setup_server};
+
+/// Regression test for RefCell borrow panic in build_now_dependencies.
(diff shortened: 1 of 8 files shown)


server/tests/test_borrow_panic.rs
```diff
@@ -0,0 +1,117 @@
...
+/// iterating its dependencies and calling build_now recursively. If the recursive
+/// call chain reached eval_symbols_from_import_stmt, which tries to borrow_mut
+/// the same symbol via add_dependency, it panicked with "RefCell already borrowed".
+///
+/// The fix: collect dependencies into an owned Vec, drop the borrow, then iterate.
+///
+/// Setup: file_b has class ClassB(file_c.ClassC), creating an ARCH_EVAL dep on
+/// file_c. file_c imports file_b. After initial build, we manually set both files
+/// to PENDING for ARCH_EVAL and call build_now(file_b) directly, bypassing
+/// process_rebuilds' dependency-aware ordering.
```
mmahrouss, line 25: What I am not sure about, if in the test we had to override the process rebuilds behavior, how was it possible actually when the program is running? 

It was coming from delayed changes, so perhaps there was too many edits, and then the dependencies where not in place yet?

fda-odoo (approved):

PR #620, opened 2026-06-04
[FIX] Fix import errors

mmahrouss:
This avoid level is too high errors in untitled files


mmahrouss:
found another bug, will update PR with a more holistic fix

cammarosano (commented):
Good job, but...

Image

fda-odoo (approved):

fda-odoo:
@mmahrouss Can you add some explanation in the test about why you are adding test data, but no test?

mmahrouss:
> @mmahrouss Can you add some explanation in the test about why you are adding test data, but no test?

Updated :)

PR #622, opened 2026-06-05
[FIX] fix paths in tests

fda-odoo:
testing path was always done with sanitize() in tests, but configs are using canonicalize, which can change the path, invalidating the tests


fda-odoo:
@mmahrouss Let me know what you tink of this pr

mmahrouss (approved):
Looks good! consistent with internal logic!

PR #624, opened 2026-06-11
Clean dependencies

cammarosano:


server/src/core/python_arch_builder.rs
cammarosano, line 1: Optionally, if we want to leave the structure for future Error handling but still remove the dependency on `anyhow`, we can just change the return types to `Result<(), ()>`. Then we later create suitable Error types as needed and change those Result types accordingly.

mmahrouss: Yes might be better, or we could define our own alias to Box<dyn Error> perhaps

mmahrouss (approved):

fda-odoo (commented):
good to me !

fda-odoo (approved):

PR #625, opened 2026-06-11
[FIX] String.as_slice() compile error

cammarosano:
When `default = ["debug_yarn"]` is enabled in cargo.toml (for debugging), OYarn resolves to a String, which does not have the `as_slice` method. The correct method to call here (that suits both String and Yarn) is `as_str`.


mmahrouss (approved):
Yep I noticed that, thanks for the PR!

Issue #628, opened 2026-06-17
Master changes to access rights


andg-odoo:
👋 Just sticking this here for reference: http://github.com/odoo/odoo/pull/166359
As you probably know `ir.access.rights` and `ir.rule` have been removed in `19.4` / current master in replacement with a singular `ir.access` model.

This may lead to issues with the special case for `ir_rule` where it has the `global` field which is a reserved keyword in python. Currently that is from 15.0 to 999 version.

 It doesn't seem like you guys do any access rights diagnostics but still good to know for the future.

fda-odoo:
Thank you for pointing it out !
I fixed the range here: https://github.com/odoo/odoo-ls/pull/629
There is indeed actually no specific diagnostics on ir_access, so limiting the hook the the right range seems to be the only fix for now

andg-odoo:
Closed via #629

PR #629, opened 2026-06-18
[IMP] Limit ir_rule hook to 19.4

fda-odoo:
As https://github.com/odoo/odoo/pull/166359 , ir_rule doesn't not exist anymore after the 19.4 branch


fda-odoo:
Fixing https://github.com/odoo/odoo-ls/issues/628

mmahrouss (approved):

PR #632, opened 2026-06-26
[FIX] server: fix progress stuck on project switch

avar-odoo:
Set WorkDoneProgressBegin cancellable to true so the Processes tab shows a dismissable cancel button.

Exit immediately on client channel disconnect instead of joining rebuild threads which may hang, causing orphaned progress entries when switching projects or worktrees.

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -725,7 +725,7 @@ impl SyncOdoo {
             token: ProgressToken::Number(session.sync_odoo.progress_token),
             value: ProgressParamsValue::WorkDone(WorkDoneProgress::Begin(WorkDoneProgressBegin {
                 title: "Odoo: Indexing".to_string(),
-                cancellable: Some(false),
+                cancellable: Some(true),
                 message: None,
                 percentage: None,
             }))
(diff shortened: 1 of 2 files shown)


avar-odoo:
Addresses: https://github.com/odoo/odoo-pycharm/issues/26

avar-odoo:
Hi @fda-odoo @cammarosano could I please request your opinion here :)

fda-odoo:
Hello @avar-odoo ,

Thank you for suggesting this fix. Currently we were not able to reproduce the pycharm issue, so this pull request will help us analyzing this.
We are closing the javascript tasks then we'll check deeper into it for the next version !

avar-odoo:
Hi @fda-odoo,

Thanks for the update. If it helps, the issue occurs when switching projects in Pycharm. 

For example, I open `worktree1`, let OdooLS index, then if I switch to `worktree2` from the menu bar, OdooLS stays stuck on `x items remaining`. Furthermore, since there's no cancel/dismiss option it becomes a persistent process notification - which can get quite annoying :melting_face: 

If needed, I can maybe provide a screen-recording and help you debug further. Thanks again

mmahrouss:
Hello @avar-odoo ! 

Thank you for the PR, I looked into the issue and managed to reproduce it, and found a fix for it. 

As for your fix, we have 2 points as to why it is a remedy but does not fix the issue: 
1. We do not want to set cancellable, because we are not intercepting it on the server side, and actually cancelling the reporting, so the cancel here does not do what is desired, we could add it in the future perhaps, but with proper handling

2. The exit does not fix the issue, as the problem was not due to the process hanging, it was just that we were not sending progress end notification on shutdown. 

Our suggested PR is #638 

Thank you and keep us updated

PR #635, opened 2026-07-01
[IMP] Address clippy warnings

mmahrouss:
This addresses most of the warnings from running cargo clippy

Main things that changed:
&Vec<_>, &String, &PathBuf in function args change to their non owned variants &[_], &str and &Path Also handles checking is_some then calling unwrap, same with Result's  ok/err

Why? : https://effective-rust.com/clippy.html

Also now clippy is added as a pre-commit git hook, so never forget to check it (you can use `-n`/`--no-verify` when needed)

NOTE: Rebased but not working, it needs to be amended still


server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -590,12 +590,12 @@ impl SymbolTable {
...
     }
 
-    pub fn create_module_from_path(session: &mut SessionInfo, path: &PathBuf, addons: NamespaceKey) -> Option<ModuleKey> {
+    pub fn create_module_from_path(session: &mut SessionInfo, path: &Path, addons: NamespaceKey) -> Option<ModuleKey> {
         let main_entry_tree = session.sync_odoo.get_main_entry_tree(addons);
         if !(main_entry_tree == (&["odoo", "addons"], &[]) && path.join("__manifest__.py").exists()) {
             return None;
         }
-        let name = path.components().last().unwrap().as_os_str().to_str().unwrap();
+        let name = path.components().next_back().unwrap().as_os_str().to_str().unwrap();
```
cammarosano, line 723: just a note that might be helpful (not requiring changes)

`Path::components` returns a double-ended iterator, and `next_back` (which visits one element) is more efficient than `last` (visits every element in the iterator until it reaches the last)

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -756,7 +756,7 @@ impl SymbolTable {
     }
 
     /// Return all symbols before the given position that are visible in the body of this symbol.
-    pub fn get_all_visible_symbols(&self, target: SymbolKey, name_prefix: &String, position: u32) -> HashMap<OYarn, Vec<SymbolKey>> {
+    pub fn get_all_visible_symbols(&self, target: SymbolKey, name_prefix: &str, position: u32) -> HashMap<OYarn, Vec<SymbolKey>> {
```
cammarosano, line 869: Cool.
Did you check the callers of these functions in this commit if they were creating owned copies of a `&str` (or `OYarn`) to pass to this function? (copies we could avoid)

As per the commit message:
> Passing &String where &str suffices forces callers to own a String.

server/src/core/xml_validation.rs
```diff
@@ -23,7 +23,7 @@ pub struct XmlValidator {
     pub xml_symbol: XmlFileKey,
     pub is_in_main_ep: bool,
     module: ModuleKey,
-    fields_cache: HashMap<OYarn, HashMap<OYarn, Vec<(SymbolKey, Option<OYarn>)>>>,
+    fields_cache: HashMap<OYarn, MembersByName>,
```
cammarosano: didn't you change this in you PR about XML-defined models?
If so, you don't need this new type at all... (just rebase this on your PR)

server/src/features/references_xml.rs
```diff
@@ -176,11 +176,7 @@ impl XmlAstReferenceVisitor {
 
     fn visit_template(session: &mut SessionInfo<'_>, node: &Node, from_module: Option<ModuleKey>, ctxt: &mut HashMap<String, ContextValue>, results: &mut Vec<Range<usize>>, target: &ReferenceTarget) {
         for attr in node.attributes() {
-            if attr.name() == "id" {
-                if XmlAstReferenceVisitor::test_attr_as_xml_id(session.st(), &attr, from_module, target) {
-                    results.push(attr.range_value());
-                }
-            } else if attr.name() == "inherit_id" {
+            if ["id", "inherit_id"].contains(&attr.name()) {
```
cammarosano: or
```rust
if matches!(attr.name(), "id"|"inherit_id") {
```
(just a half-suggestion, I don't have a particular preference)

server/src/features/workspace_symbols.rs
```diff
@@ -49,31 +49,29 @@ impl WorkspaceSymbolFeature {
...
-        if path.is_some() && session.st().has_range(symbol) {
+        if let Some(path) = path
+        && session.st().has_range(symbol) {
             //Test if symbol should be returned
-            if string_fuzzy_contains(&session.st().name(symbol), &query) {
+            if string_fuzzy_contains(session.st().name(symbol), query) {
                 let name = session.st().name(symbol).to_string();
-                let range = session.st().range(symbol).clone();
-                WorkspaceSymbolFeature::add_symbol_to_results(session, symbol, &name, path.unwrap(), container_name.clone(), Some(&range), can_resolve_location_range, results);
+                let range = *session.st().range(symbol);
```
cammarosano, line 66: note: this has to do with `Range` being `Copy` (a type that impls the Copy trait)

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -94,7 +94,7 @@ impl SymbolTable {
     }
     pub fn add_new_function(&mut self, parent: SymbolKey, name: &str, range: &TextRange, body_start: &TextSize) -> FunctionKey {
         let is_external = self.is_external(parent);
-        let function_symbol = FunctionSymbol::new(name, parent, *range, body_start.clone(), is_external);
+        let function_symbol = FunctionSymbol::new(name, parent, *range, *body_start, is_external);
```
cammarosano: I love this commit ([IMP] clippy: remove .clone() on Copy types)
❤️ 
Bye bye `clone`s...

server/src/core/file_mgr.rs
```diff
@@ -249,16 +249,16 @@ impl FileInfo {
                 TokenKind::Comment => {
                     let text = &source[token.range()];
                     if text.starts_with("#noqa") || text.starts_with("# noqa") || text.starts_with("# odools: noqa") {
-                        let after_noqa = text.split("noqa").skip(1).next();
+                        let after_noqa = text.split("noqa").nth(1);
```
cammarosano, line 530: 🤯

server/src/core/python_arch_eval.rs
```diff
@@ -978,7 +978,7 @@ impl PythonArchEval {
             // Check for type annotation `typing.Self`, if so, return a `self` evaluation
             // And give it priority over other evaluations
             if evaluations.iter().any(|evaluation|
-                SymbolTable::follow_ref(
+                !SymbolTable::follow_ref(
```
cammarosano, line 980: ~what?? (Maybe the rest of the diff is in a different commit...)~
Note for someone who might be reviewing by commits: the change of `len() > 0` to `is_empty()` is in some other commit.

server/src/core/config.rs
```diff
@@ -941,7 +941,7 @@ fn read_config_from_file<P: AsRef<Path>>(path: P) -> Result<HashMap<String, Conf
         });
 
         // python_path
-        entry.python_path.as_mut().map(|sourced| { sourced.sources.insert(path.sanitize());});
```
cammarosano: that was crazy. running map and discarding the result, just for the side-effects.

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -688,7 +685,7 @@ impl SymbolTable {
                 if iter_sym.len() > 1 {
                     trace!("TODO: explore all implementation possibilities");
                 }
-                for fk in symbol_tree_content[1..symbol_tree_content.len()].iter() {
+                if let Some(fk) = symbol_tree_content[1..symbol_tree_content.len()].iter().next() {
                     let _iter_sym = self.get_sub_symbol(iter_sym[0], fk.as_ref(), position);
                     iter_sym = _iter_sym.symbols;
                     // TODO: this is a loop that runs only once! Fix me!
```
cammarosano: maybe move this comment to above the `if let Some() ... .iter().next()` that replaced the loop?
Unless you're sure this is not an error/bug and it's no longer a TODO...

server/src/core/config.rs
```diff
@@ -1107,7 +1107,7 @@ fn apply_extends(
     let mut visited = HashSet::default();
     for parent in nodes.iter().filter(|(_, n)| n.parent.is_none()){
         let mut stack = vec![parent.0.clone()];
-        while let Some(current) = stack.pop().map(|value| {ordered_nodes.push(value.clone()); value}) {
+        while let Some(current) = stack.pop().inspect(|value| {ordered_nodes.push(value.clone());}) {
```
cammarosano: if this code is going to be replaced, ok.
Otherwise, we should call `ordered_nodes.push(current.clone())` on the line below.

server/src/core/file_mgr.rs
```diff
@@ -340,7 +340,7 @@ impl FileInfo {
                     !filter.paths.iter().any(|p| p.matches(&self.uri))
                 }
             }
-        }).collect::<Vec<_>>();
+        }).cloned().collect::<Vec<_>>();
```
cammarosano, line 615: ??
Isn't this changing the type of `diagnostic_filters`?
What issue does clippy solve here?

server/src/core/file_mgr.rs
```diff
@@ -479,7 +479,13 @@ impl FileInfo {
 pub struct FileMgr {
     pub files: HashMap<String, Rc<RefCell<FileInfo>>>,
     untitled_files: HashMap<String, Rc<RefCell<FileInfo>>>, // key: untitled URI or unique name
-    workspace_folders: HashSet<(String, Uri)>,
+    workspace_folders: HashSet<(String, String)>, // (name, uri string) - stored as String to avoid interior mutability in HashSet keys (Uri)
```
cammarosano: I guess this refers to https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#mutable_key_type.

Did you check if this is not a false-positive? `lsp-types::Uri` seems to fit the description...
>  False Negatives
This lint does not follow raw pointers (*const T or *mut T) as Hash and Ord apply only to the address of the contained value. This can cause false negatives for custom collections that use raw pointers internally.

> Configuration
ignore-interior-mutability: A list of paths to types that should be treated as if they do not contain interior mutability
(default: ["bytes::Bytes"])

server/src/core/odoo.rs
```diff
@@ -1935,14 +1935,14 @@ impl Odoo {
                             let tree_path = path.to_tree_path();
                             if tree.is_none() ||
                             (session.st().get_symbol(session.sync_odoo.get_main_entry().borrow().root.into(), tree.as_ref().unwrap().as_slice(), u32::MAX).is_empty()
-                            && session.sync_odoo.get_main_entry().borrow().data_symbols.get(&path.sanitize()).is_none())
+                            && !session.sync_odoo.get_main_entry().borrow().data_symbols.contains_key(&path.sanitize()))
```
cammarosano: I know this is automated...
But you can use `sanitize_cow` here

server/src/features/completion.rs
```diff
@@ -61,7 +61,7 @@ impl CompletionFeature {
         let file_info_ast = file_info.borrow().file_info_ast.clone();
         let file_info_ast = file_info_ast.borrow();
         let ast = file_info_ast.get_stmts().unwrap();
-        complete_vec_stmt(ast, session, file_symbol, offset).or_else(|| complete_name(session, file_symbol, offset, false, &S!("")))
```
cammarosano: 🤣

server/src/utils.rs
```diff
@@ -117,19 +117,19 @@ pub fn is_symlink_cs(path: String) -> bool {
...
         let s = self.as_str();
         // Detect legacy UNC path (file:////)
         if s.starts_with("file:////") {
             legacy_unc_paths().store(true, Ordering::Relaxed);
         }
         let str_repr = s.replace("file:////", "file://");
-        let url = url::Url::from_str(&str_repr).map_err(|_| ())?;
-        url.to_file_path()
+        let url = url::Url::from_str(&str_repr).map_err(|e| Box::new(e) as Box<dyn std::error::Error>)?;
+        url.to_file_path().map_err(|_| Box::<dyn std::error::Error>::from("Failed to convert URL to file path"))
```
cammarosano, line 134: Maybe simpler to just use `Result<PathBuf, String>` here

```suggestion
    fn to_file_path(&self) -> Result<PathBuf, String>;
}

impl ToFilePath for lsp_types::Uri {
    fn to_file_path(&self) -> Result<PathBuf, String> {
        let s = self.as_str();
        // Detect legacy UNC path (file:////)
        if s.starts_with("file:////") {
            legacy_unc_paths().store(true, Ordering::Relaxed);
        }
        let str_repr = s.replace("file:////", "file://");
        let url = url::Url::from_str(&str_repr).map_err(|e| e.into())?; // or maybe map_err(String::from). I'm assuming e has String representation (they usually do)
        url.to_file_path().map_err(|_| "Failed to convert URL to file path".to_string()))
```

server/tests/test_basics.rs
```diff
@@ -31,6 +31,9 @@ fn test_custom_entry_point() {
 
 
 #[test]
+// The parsed source contains the float literal `3.14`; comparing against it is
```
cammarosano: couldn't you just use 6.66 in the tests instead? 😄

server/src/core/symbols/storage/lifecycle.rs
```diff
@@ -246,7 +246,7 @@ impl SymbolTable {
         let ns = &mut self.namespaces[parent];
         let best = ns.directories.iter()
             .enumerate()
-            .filter(|(_, dir)| PathBuf::from(path).starts_with(&dir.path))
+            .filter(|(_, dir)| Path::new(path).starts_with(&dir.path))
```
cammarosano, line 261: this commit (use Path when possible) is beautiful! ❤️

.git-blame-ignore-revs
```diff
@@ -0,0 +1,66 @@
+# Revisions to ignore in `git blame`.
```
cammarosano, line 1: cool!
just remember to update this file if this gets rebased...

cammarosano (approved):
Very nice! Clippy for president! (mdms for vice)

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: nice, thanks for highlighting that!

server/src/core/xml_validation.rs
mmahrouss: yes you are right, but then both PRs are not approved, I am not sure which one should I rebase on which, so I will keep the comment here to change the code whenever one of them is merged

server/src/features/references_xml.rs
mmahrouss: Hmmm I agree `matches!` here works better than `contains` on an array, but I believe we have that in multiple code sites. 
I think I will just change it only here, and use it later when faced with the same pattern

server/src/core/python_arch_eval.rs
mmahrouss: okay that is weird, Maybe I should change the commit history here

server/src/core/config.rs
mmahrouss: I agree

server/src/core/file_mgr.rs
mmahrouss: It seems to be movingthe cloning of the values in the iterator from before the filter to after the filter.  
it is not changing the type because there was a clone earlier in the function already. 

the warning it is addressing is https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#iter_overeager_cloned 

So it aims  to only clone at the end to minimize copies

server/src/core/file_mgr.rs
mmahrouss: You are right, I did misunderstand interior mutability, I will update the config and undo that change

.git-blame-ignore-revs
mmahrouss: yes for sure, and we need to merge the PR in the CLI because github changes the SHAs when we merge from its UI

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: it is only used once at `get_all_inferred_names` which just relays a `&str` from the fn args

server/src/core/python_arch_eval.rs
mmahrouss: Done!

server/src/core/odoo.rs
mmahrouss: Yes it was done in a later commit!

server/src/core/file_mgr.rs
cammarosano: ah, ok, I couldn't see the first `cloned` being removed when I did the review per commit. It's clear when seeing the aggregate diff. My bad.

server/src/core/odoo.rs
cammarosano: oops :grimacing:

server/src/utils.rs
mmahrouss: I chose to keep it but drop the unneeded map_err

fda-odoo:
merged in the rebased version of this PR: https://github.com/odoo/odoo-ls/pull/656

PR #636, opened 2026-07-02
Completion and diagnostics for the new access search domain operator

mmahrouss:


server/src/core/evaluation.rs
```diff
@@ -1928,41 +1963,45 @@ impl Evaluation {
                             };
                             if ["date", "datetime"].contains(&ttype.as_str()) {
                                 date_mode = true;
+                            } else if index == split_expr.len() - 1
+                                && access_op
+                                && matches!(ttype.as_str(), "many2one") // ID has type integer, so we cannot get a strict check easily
```
fda-odoo: but so, ID will not validate `matches!(ttype.as_str(), "many2one")`, right?

fda-odoo (changes requested):
nice job :)

server/src/core/evaluation.rs
mmahrouss: Yes you are right, I did first write the code to check if we found an invalid one, now that it is reversed to check for a valid one, I should change the condition. Okay let's not merge it till I fix that and add a test. Thank you for pointing it out:)

fda-odoo (approved):

PR #637, opened 2026-07-03
[FIX] Fix and refactor SymbolTable::get_symbol

mmahrouss:
before there was bag in the last loop that was early returning, effectively only processing content paths with length 2 only. Which was okay because all use cases happen to only pass at most 2 content elements

Also multiple symbols with the same name where not handled, here it is also handled by the `flat_map` in the second loop. That was the second TODO here

This PR uses correct loops, and combines the checks and loops into concise declarative maps. The method was also formatted

### DIFF
--- a/server/src/core/symbols/symbol_table_impl.rs
+++ b/server/src/core/symbols/symbol_table_impl.rs
@@ -2145,3 +2145,235 @@ impl SymbolTable {
     }
 
 }
+
+#[cfg(test)]
+mod get_symbol_tests {
+    use super::*;
+    use crate::core::entry_point::{EntryPoint, EntryPointType};
+    use ruff_text_size::TextSize;
+
+    /// Build an empty symbol table with a single root and return `(table, root_key)`.
+    fn empty_table_with_root() -> (SymbolTable, SymbolKey) {
+        let mut st = SymbolTable::new();
+        let entry = EntryPoint::new(
+            &mut st,
+            "/test".to_string(),
+            vec![],
+            EntryPointType::MAIN,
+            None,
+            None,
+        );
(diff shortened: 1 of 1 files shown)


server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -645,59 +645,44 @@ impl SymbolTable {
...
-                    let _iter_sym = self.get_sub_symbol(iter_sym[0], fk.as_ref(), position);
-                    iter_sym = _iter_sym.symbols;
-                    if iter_sym.is_empty() {
-                        return vec![];
-                    }
-                }
-            }
-        } else {
-            if symbol_tree_content.len() == 0 {
+    pub fn get_symbol<S: AsRef<str>>(
```
cammarosano, line 653: this is amazing! 🤯 

If I were you: I would write UNIT tests for this.
Yes, unit: same file, with a helper to build a test symbol table with some parent-children relationships. And then write the test-case scenarios (just file, file + 1 symbol content, file + nested symbol content, symbols with same name...).
Maybe it's just simpler to do the usual integration test (and have a test python file),... note that you don't need a full odoo for this one.
Ok, that seems a bit complicated. Just follow your heart if you wanna burn that much tokens! 😄

cammarosano (approved):
very cool!!

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: I did that, can you take a look :) ?

cammarosano: I love it! :heart_on_fire:

fda-odoo (approved):
coool! that's much more clean :D I like it !

PR #638, opened 2026-07-09
[FIX] Terminate progress reporting

mmahrouss:
In process_rebuilds, progress was not ended when a termination signal is received. The fix comes in two stages
1. Send an End notification in process rebuilds
2. Make sure any pending notifications are flushed after shutdown in the master process

Closes https://github.com/odoo/odoo-pycharm/issues/26

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -270,6 +270,25 @@ impl Server {
         info!(message);
     }
 
+    /// Drain pending respones and notifications from worker threads to the client.
+    fn flush_pending_messages(receivers_w_to_s: &[Receiver<Message>], connection: &Connection) {
+        for receiver in receivers_w_to_s.iter() {
+            while let Ok(msg) = receiver.try_recv() {
+                match msg {
+                    Message::Response(r) => {
+                        let _ = connection.sender.send(Message::Response(r));
+                    },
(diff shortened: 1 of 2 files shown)


server/src/server.rs
```diff
@@ -270,6 +270,28 @@ impl Server {
...
+     * has broken out).
+     * Only notifications are flushed: requests are dropped (e.g. a queued WorkDoneProgressCreate
+     * would just create a new, unclosable progress on the client) and responses are dropped too
+     * (nothing is waiting for them on a connection that's already being shut down). The internal
+     * Shutdown sentinel notification is also dropped, as it's never meant to reach the client.
+     */
+    fn flush_pending_notifications(receivers_w_to_s: &[Receiver<Message>], connection: &Connection) {
+        for receiver in receivers_w_to_s.iter() {
+            while let Ok(msg) = receiver.try_recv() {
+                if let Message::Notification(n) = msg
```
fda-odoo: I agree to not send requests, but maybe we could send responses?

fda-odoo (changes requested):

server/src/server.rs
mmahrouss: Updated, can you recheck please?

cammarosano (approved):
👌

fda-odoo (approved):

PR #640, opened 2026-07-11
[FIX] server: no OLS03023 when the inverse field is overridden

MiquelRForgeFlow:
The inverse field of a One2many may be declared several times, e.g. when a model overrides the field of an inherited abstract model to point to the concrete model instead. As the last declaration  overrides the comodel of the previous ones at runtime, only raise OLS03023 when no declaration of the inverse field points to the current model.

(If you prefer me to target `release` or another branch, just tell me)

### DIFF
--- a/server/src/core/python_validator.rs
+++ b/server/src/core/python_validator.rs
@@ -575,13 +575,24 @@ impl PythonValidator {
                                     comodel_eval_weaks.extend(followed);
                                 }
                             }
+                            // The inverse field may be declared several times, e.g. when a
+                            // model overrides the field of an inherited abstract model to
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
Hi fda-odoo 😊 when you have a bit if free time, please check this.

fda-odoo:
It looks good to me!
Alpha seems a valid target for this PR, so it can be merged in next 1.4.0 version.
thank you for suggesting the fix !

mmahrouss (approved):
Looks good to me, thank you:) and thanks for including a test too!

PR #642, opened 2026-07-13
JS features via virtual docs

cammarosano:
There is useful information in server/docs/owl-virtual-docs.md.

Intermediate JS/OWL* commits do not compile, they were split for easier review.


server/Cargo.toml
```diff
@@ -67,5 +67,5 @@ inherits = "release"
 debug = true
 
 [profile.release]
-codegen-units = 1
-lto = "thin"
+# codegen-units = 1
```
fda-odoo: don't forget to remove this :)

server/src/core/odoo.rs
```diff
@@ -1907,8 +1907,8 @@ impl Odoo {
         }
         let path = match params.text_document.uri.scheme().map(|scheme| scheme.to_lowercase()) {
             Some(schema) if schema == "file" => {
-                let uri = params.text_document.uri.to_string();
-                if !uri.ends_with(".py") {
+                let uri = params.text_document.uri.as_str();
+                if [".py", ".js", ".ts"].iter().all(|&ext| !uri.ends_with(ext) ) {
```
fda-odoo: that's not js-related, but maybe .pyi could fit here too

server/src/features/semantic_tokens.rs
```diff
@@ -162,6 +162,35 @@ impl SemanticTokensFeature {
...
+        let content = {
+            let fi = file_info.borrow();
+            let fia = fi.file_info_ast.borrow();
+            match fia.text_document.as_ref() {
+                Some(td) => td.contents().to_string(),
+                None => return SemanticTokens { result_id: None, data: vec![] },
+            }
+        };
+
+        // tsserver is the only source of JS semantics. Without it (CLI / most tests)
```
fda-odoo: we'll have to provide tsserver in CLI too (but probably for 1.5.1)

server/src/core/file_mgr.rs
```diff
@@ -132,17 +133,26 @@ impl PythonAst {
...
     /// Positions of OWL `static template = "some.xml_id"` string literals found in this JS file.
-    /// Each entry is (LSP Range of the string content, xml_id value, enclosing class name).
+    /// Each entry is (byte range of the string content, xml_id value, enclosing class name).
+    /// The range is converted to LSP coordinates by consumers.
     pub js_template_refs: Vec<JsTemplateRef>,
     /// Component descriptors extracted from OXC analysis of this JS file.
     pub js_component_descriptors: Vec<ComponentDescriptor>,
+    /// Every module specifier this JS file imports from, verbatim as written (incl.
+    /// bare `import "x"` and `export … from`). Sorted and deduplicated.
+    pub js_imports: Vec<String>,
```
fda-odoo, line 145: I'm not sure it's worse because imports string are most of the time not that small, but as we cache these values, OYarn could maybe fit here

server/src/core/text_document.rs
```diff
@@ -49,6 +49,41 @@ impl RangeExt for lsp_types::Range {
 }
 
 
+/// Convert a byte offset in `text` to an LSP position, using a prebuilt line `index` for
+/// `text`. Offsets past the end of `text` are clamped to it. For content that has no
+/// [`TextDocument`] (e.g. the OWL virtual docs, or files read straight from disk).
+pub(crate) fn offset_to_position(
```
fda-odoo: this should not be here. all translations for ranges, offset, position, ... are in file_mgr. Moreover this file is code coming from ruff (see first lines of the file), and should be kept nearly as it is

server/src/features/owl_virtual.rs
```diff
@@ -0,0 +1,1453 @@
...
+    /// Path of the shim (`<stem>.__ols_shim__.js`), sibling of the real file.
+    pub(crate) path: String,
+    /// The shim content (`real` + `\nexport { … };`).
+    content: String,
+}
+
+/// A built virtual document for one OWL **component**: import line + one `@this` function
+/// per template expression, plus the data to map results back. Owned (independent of
+/// `session`), so it can be held across `&mut session` bridge calls.
+pub(crate) struct OwlVirtualDoc {
```
fda-odoo, line 155: This OwlVirtualDoc is duplicating logic of file_mgr and its text_document.
Access, caching, updates, offset translation of any file in cache should be done at the same place.
You are duplicating here all the 'position_to_offset', 'real_byte_to_position', etc... and moving some cache management outside the file_mgr.
I would try to move that in file_mgr, maybe by making TextDocument an enum, or explicitly handling virtual files in file_mgr (we already have a logic for files that don't have a path to the disk for 'untitled' files)
You have some variables here that should not be on the regular TextDocument (like local_decls), but then I think you should still separate content caching and its maangement to augmented data linked to it.
Let me know what you think about it. Maybe waiting for tsgo would ease the process regarding encoding

server/src/features/owl_virtual.rs
```diff
@@ -0,0 +1,1453 @@
...
+/// The real `.js` a shim was copied from (`<stem>.__ols_shim__.js` → `<stem>.js`).
+pub(crate) fn shim_to_real(shim_path: &str) -> String {
+    let stem = shim_path.strip_suffix(SHIM_JS_SUFFIX).unwrap_or(shim_path);
+    format!("{stem}.js")
+}
+
+/// Semantic tokens for an XML OWL-template file: template-name tokens (in-house, works
+/// without tsserver) merged with JS-expression tokens (delegated to tsserver via the
+/// virtual docs). `None` when neither source produced a token.
+pub fn semantic_tokens_xml(session: &mut SessionInfo, file_info: &Rc<RefCell<FileInfo>>) -> Option<SemanticTokens> {
```
fda-odoo, line 242: why semantic_tokens_xml is in this file and not in semantic_token file?

server/src/features/owl_virtual.rs
```diff
@@ -0,0 +1,1453 @@
...
+    let mut raw_tokens: Vec<(Range, u32, u32)> = vec![];
+
+    // 1. Template-name tokens. Independent of tsserver: we only need to parse the XML.
+    let xml = {
+        let fi = file_info.borrow();
+        let fia = fi.file_info_ast.borrow();
+        fia.text_document.as_ref().map(|td| td.contents().to_string())
+    };
+    if let Some(xml) = xml.as_deref() {
+        if let Ok(document) = roxmltree::Document::parse(xml) {
```
fda-odoo, line 253: Maybe for later, but that's again a new flow where we have to parse the xml content again. We already do if for regular parsing and DocumentSymbol.
Maybe we could try to compute DocumentSymbol and semanticTokens together, then serve the cached value to the second request, that most of the time will arrive at the same time (file opening).
Just an idea though, probably for a next version. Ast parsing is still fast anyway

server/src/features/owl_virtual.rs
```diff
@@ -0,0 +1,1453 @@
...
+use crate::core::tsserver_bridge::{ts_to_lsp_location, TsLocation, TsServerBridge};
+use crate::features::owl_expr::{compile_owl_expr, interp_chunk_ranges, this_token_at};
+use crate::features::owl_xml_utils::{TEMPLATE_NAME_ATTRS, component_tag_name_range, is_owl_expression_attr, is_owl_interpolation_attr, is_prop_expr_attr, tag_is_component};
+use crate::features::semantic_tokens::{SemanticTokensFeature, TokMod, TokType, U16ToByte};
+use crate::threads::SessionInfo;
+use crate::utils::HashMap;
+
+/// A template-local in scope for an expression (`t-set` or `t-foreach`/`t-as`), emitted
+/// into the function preamble as a `let` so tsserver types bare identifiers. `decl_offset`
+/// is the XML byte of the declared name, for go-to-definition on a local.
```
fda-odoo, line 35: I had to read that 20 times :')

server/src/core/tsserver_bridge.rs
```diff
@@ -590,6 +767,49 @@ impl TsServerBridge {
...
+    /// because our legend's first slots mirror tsserver's numbering — so no translation is
+    /// needed here beyond splitting tsserver's packed classification value.
+    pub fn get_semantic_tokens(&mut self, file_path: &str) -> Vec<(u32, u32, u32, u32)> {
+        // A length past EOF classifies the whole file; tsserver clamps to the real end.
+        let request_seq = match self.send_request(
+            "encodedSemanticClassifications-full",
+            json!({ "file": file_path, "start": 0, "length": i32::MAX, "format": "2020" }),
+        ) {
+            Ok(seq) => seq,
+            Err(_) => return vec![],
```
mmahrouss, line 807: Perhaps here and in other similar places, we can at least log, if not propagate the error

server/src/core/odoo.rs
mmahrouss: This is from the other pr, but since you defined `let uri = params.text_document.uri.as_str();` you could also use it here: 
```suggestion
                            message: format!("Invalid file URI: {}", uri),
```

server/src/core/odoo.rs
```diff
@@ -1907,8 +1915,8 @@ impl Odoo {
         }
         let path = match params.text_document.uri.scheme().map(|scheme| scheme.to_lowercase()) {
             Some(schema) if schema == "file" => {
-                let uri = params.text_document.uri.to_string();
-                if !uri.ends_with(".py") {
+                let uri = params.text_document.uri.as_str();
+                if [".py", ".js", ".ts", ".xml"].iter().all(|&ext| !uri.ends_with(ext) ) {
                     return Ok(None);
                 }
```
mmahrouss, line 1906: Maybe we can do this check after `to_file_path` and then use `extension()` and check with that? so we get the robustness of the `std::Path(Buf)` methods?

server/src/core/odoo.rs
```diff
@@ -1887,7 +1888,13 @@ impl Odoo {
                         }
                     },
                     Ast::XmlAst => {
-                        return Ok(HoverFeature::hover_xml(session, file_symbol, &file_info, params.text_document_position_params.position.line, params.text_document_position_params.position.character));
+                        let Position { line, character } = params.text_document_position_params.position;
+                        // OWL-template JS expressions first; everything else → XML hover.
+                        // @todo: check if not breaking python-related hover
+                        if let Some(hover) = owl_virtual::hover_xml_owl(session, &file_info, line, character) {
```
fda-odoo, line 1880: maybe we could trigger hover_xml_owl only for files that are loaded as web_assets (actually this information is on xml templates symbols, but we should store it on xml_file_symbol too)
It would avoid all the unnecessary parsing

server/src/core/tsserver_bridge.rs
```diff
@@ -590,6 +767,49 @@ impl TsServerBridge {
...
+            .chunks_exact(3)
+            .filter_map(|triple| {
+                let start = triple[0].as_u64()? as u32;
+                let length = triple[1].as_u64()? as u32;
+                let classification = triple[2].as_u64()? as u32;
+                let token_type = (classification >> 8).checked_sub(1)?;
+                let modifiers = classification & 0xFF;
+                Some((start, length, token_type, modifiers))
+            })
+            .collect()
```
mmahrouss, line 835: From the docstring of chunks_exact: 
> If your `chunk_size` is a constant, consider using [`as_chunks`] instead, which will
    give references to arrays of exactly that length, rather than slices.


```suggestion
        let (chunks, _remainder) = spans
            .as_chunks::<3>();
        chunks
            .into_iter()
            .filter_map(|triple| {
                let start = triple[0].as_u64()? as u32;
                let length = triple[1].as_u64()? as u32;
                let classification = triple[2].as_u64()? as u32;
                let token_type = (classification >> 8).checked_sub(1)?;
                let modifiers = classification & 0xFF;
                Some((start, length, token_type, modifiers))
            })
            .collect()
```

server/src/features/owl_component_utils.rs
```diff
@@ -0,0 +1,136 @@
...
+/// The base-most of the classes declaring one template, skipping any whose descriptor is gone.
+/// Runs at query time, when every file is built and the super-chains are therefore complete.
+fn resolve_component(
+    classes: &[String],
+    descriptors: &HashMap<String, ComponentDescriptor>,
+) -> Option<String> {
+    let mut candidates: Vec<&String> = classes
+        .iter()
+        .filter(|class_name| descriptors.contains_key(*class_name))
+        .collect();
```
mmahrouss: As discussed, currently `contains_key` will always be true, because the `js_component_by_template` and `component_descriptors` are created together in build, and neither get cleaned, thus, they cannot go out of sync. So we can drop the check or add a debug assertion, and we should consider actually cleaning them up on dropping files

server/src/features/owl_virtual.rs
fda-odoo: More generally, we had 'definition.rs' file, 'declaration.rs' file, etc... one for each feature, but it was always for python. Now we have owl_virtual that has some, semantic_token_xml is in the end only handling python,... 
I think we should refactor all these files to reorganize better per language (in a future task)

fda-odoo (commented):
First pass that analyze the global structure and analyze of the code. I didn't go deeply in algorithm exactness, but focused on the global vision.
I like the idea and the result from a user point of view is amazing ! As a maintainer however, I think it can still be improved. Some part of the code should be refactored and reorganized for more clarity.
I think however that this refactoring will be mandatory for the upgrade to tsgo (typescript 7), so maybe we can wait for that to move everything cleanly to the next step

server/src/core/js_arch_builder.rs
```diff
@@ -1,74 +1,78 @@
...
+///
+/// Variables are deliberately absent — the Python walk skips them too, and they drown the
+/// picker. A top-level `const` bound to a function or a class expression is not a variable in
+/// that sense and IS collected.
 #[derive(Debug, Clone)]
-pub struct ComponentMember {
-    pub name: String,
-    pub kind: MemberKind,
-    pub typ: InferredType,
+pub struct JsDecl {
```
mmahrouss: I do not feel like Decl is a nice abbreviation of Declaration, maybe just use `JsDeclaration` it is not that big for a type name anyway

cammarosano:
Force pushes are mostly rebases after the target branch (semantic tokens python) changed.

The only new code is the last commit: [[IMP] add signature/docs to completion suggestions](https://github.com/odoo/odoo-ls/pull/642/commits/7294fad02a60827a6f2071c07ec755f0b302606f)

cammarosano:
@fda-odoo 2 new commits: they solve major issues:
- now we don't need to tell the user to remove/rename their jsconfig.json
- now all the type declarations in .d.ts files (meant for... tooling!) are properly included in the project, and type resolutions work as it should.

server/src/core/text_document.rs
mmahrouss: Since this file is a reference to the ruf file we are reusing, check the first comment. I would rather put these functions somewhere else, perhaps file_mgr or utils

mmahrouss (commented):
Review part 1. I still did not check everything

fda-odoo:
Two lasts commits are good additions! 
I let you address Mahrous comments, then I think we can merge that. My comments should be delayed for the tsgo refactoring.
Let's include that in the 1.5.0, and we'll plan together (with mahrous too ofc) the next steps for the JS.
Congrats on making this virtual document feature, that was a big research field !

fda-odoo (approved):

server/src/core/odoo.rs
cammarosano: agreed, but in a another PR.
This pattern is repeated in handle_hover, handle_gotos, etc. So we would change all of them (and maybe extract code into a little helper).
Not doing this now to avoid a big post-review diff.

server/src/features/owl_virtual.rs
cammarosano: Yeah, agreed.
But one of the reasons to manage it here is because FileMgr + TextDocument is overkill: we don't expect to receive different versions of a file from the client and have to handle that. We manage our virtual doc entirely.
I guess this virtual document could somehow refer to the existing (real) TextDocument, instead of holding fields like `real_content` and `real_index`...
Anyway, I agree in principle that File manager could be refactor to also hold virtual documents. For later...

mmahrouss (approved):

server/src/features/owl_virtual.rs
cammarosano: > why semantic_tokens_xml is in this file and not in semantic_token file?

My bad. In a first moment, it only had the part that depended on virtual doc (now the 2nd block), so it has virtual doc-powered feature. Later came the template-name tokens.
I'll leave this reorganization to the future task you mentioned in the second comment

server/src/features/owl_virtual.rs
cammarosano: the honest answer: because I did not know we offered DocumentSymbol for xml files 😄 

But now that I'm thinking about it: they are independent features, and semantic tokens are not always enabled (and thus not always asked for).
And yes: xml parsing is light/fast.
To be honest, I have not worried too much about performance of feature requests (as opposed to build/re-build).

PR #651, opened 2026-07-15
semantic tokens python

cammarosano:


andg-odoo:
I've tested this locally when testing #642 and looking at the difference of the two it seems the same implementation. Works great! Excited to see this grow and expand on more parts like domain strings and other fancy odoo-specific concepts.

I only noticed two issues with it which seems to be known but though I'd point it out anyway.

**The pattern:** at a couple of sites we emit a token that carries *less* information than the TextMate/treesitter grammar already had, and since the LSP token wins on priority (125 > 100), the result colors *worse* than with the feature off.

**1. Field/attribute declaration lines is a plain `variable`.** `classify()`'s `Variable` arm maps `TokenOrigin::Name` to `variable`, and the visitor sees the LHS of `field = fields.Char(...)` as a bare `Expr::Name`. So the declaration line loses the grammar's `@variable.member` and reads as a plain variable, even though the *usage* sites (`self.field`, origin `Attr`) correctly get `property`:

651_member_var_downgrade

I see this is already the "dedicated Odoo-field detection" item under *Deliberately deferred* mostly noting that until it lands, declarations look slightly worse than with no server tokens.

**2. Module/class-level constants lead to plain `variable`.** ALL_CAPS names like `API_VERSION` lose the grammar's `@constant`. The Python path only ever sets `Static`/`DefaultLibrary`, so nothing carries the "constant" signal. `READONLY` is already a reserved slot in the legend, so this is just OR-ing the bit for ALL_CAPS / `Final`-annotated names in the `Variable` arm,no legend change needed:

651_constant_downgrade

Let me know if it's better to write individual issues for these as it merges! Thanks as per usual for all the great work, loving the recent improvements.

server/src/server.rs
```diff
@@ -431,7 +440,7 @@ impl Server {
                     ResolveCompletionItem::METHOD => {
                         info!("Got ignored CompletionItem/resolve")
                     }
-                    _ => {panic!("Not handled Request Id: {}", r.method)}
+                    _ => {warn!("Not handled Request Id: {}", r.method)}
```
fda-odoo: We should keep the panic here. Or at least check that the method contains a '$' as first character to log the issue as we do for notifications.
But if the message is not optional, we should consider that we are in an invalid state, and so panic, as client is expecting a mandatory answer

server/src/features/semantic_tokens.rs
```diff
@@ -0,0 +1,340 @@
...
+use crate::core::evaluation::{Evaluation, EvaluationSymbolPtr, ExprOrIdent};
+use crate::core::file_mgr::FileInfo;
+use crate::core::symbols::symbol_keys::{SourceFileKey, SymbolKey};
+use crate::features::ast_utils::AstUtils;
+use crate::threads::SessionInfo;
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// Single source of truth for token *type* indices. The order here MUST match the
+/// order of `legend().token_types` so the `u32` we send to the client lines up.
```
fda-odoo, line 19: I tend to prefer data structure that enforce this kind of condition.
what about implementing
```TokType::as_token() -> SemanticTokenType```
then return the legend with
```rust
SemanticTokensLegend {
            token_types: vec![
                TokType::Namespace::as_token(),
                TokType::Class::as_token(),
                TokType::Function::as_token(),
                ...
            ],
```

server/src/features/semantic_tokens.rs
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    pub fn tokens_python(session: &mut SessionInfo, file_symbol: SourceFileKey, file_info: &Rc<RefCell<FileInfo>>) -> SemanticTokens {
+        // Raw, unsorted tokens: (range, token type index, modifier bitset). We
+        // collect first, then sort by (line, char) and delta-encode at the end.
+        let mut raw: Vec<(Range, u32, u32)> = vec![];
+
+        // Borrow the AST for the whole walk. `session` is a separate object, and
+        // `get_symbol_from_expr` does not need this borrow, so there is no conflict.
```
fda-odoo: that's AI reassuring itself, I think you can remove this comment :)

fda-odoo (commented):
small comments I got while reading that (not a review)

server/src/server.rs
cammarosano: reverted

cammarosano:
@andg-odoo Thanks for your feedback! Give it a go now...

andg-odoo:
image
image
image
image


Houston we have color, works great, it's nice to see domains colored and methods and everything in computes.

mmahrouss (approved):
Amazingly cool!

server/src/features/goto_utils.rs
```diff
@@ -48,157 +46,53 @@ pub struct GotoSource {
 pub struct GotoUtils {}
 
 impl GotoUtils {
-    fn check_for_domain_field(session: &mut SessionInfo, eval: &Evaluation, file_symbol: SourceFileKey, call_expr: &Option<ExprCall>, offset: usize, sources: &mut Vec<GotoSource>) -> bool {
-        let (field_name, field_range) = if let Some(eval_value) = eval.value.as_ref() {
-            if let Some(expr) = eval_value.as_string_literal() {
-                (expr.value.to_str(), expr.range)
-            } else {
-                return false;
+    fn push_string_sources(session: &mut SessionInfo, resolution: StringResolution, file_symbol: SourceFileKey, string_range: TextRange, sources: &mut Vec<GotoSource>) {
```
fda-odoo, line 49: Nice refactoring of all the gotoutils methods !

fda-odoo (approved):

PR #652, opened 2026-07-16
[IMP] Add config toggle to disable Javascript

mmahrouss:
Maybe the config can be renamed, I am up for suggestions

### DIFF
--- a/server/src/core/config/config_key_spec.rs
+++ b/server/src/core/config/config_key_spec.rs
@@ -32,6 +32,7 @@ pub enum ConfigKey {
     DiagnosticFilters,
     NoTypeshedStubs,
     TsServerCommand,
+    DisableJavascript,
 }
 
 pub(super) fn specs() -> Vec<FieldSpec> {
@@ -95,6 +96,12 @@ pub(super) fn specs() -> Vec<FieldSpec> {
             .triggers_restart(),
         FieldSpec::new(TsServerCommand, "tsserver_command", Str) //it could maybe be nice to resolve config variables like {workspaceFolder}
             .triggers_restart()
-            .default(ConfigValue::str("tsserver"))
+            .default(ConfigValue::str("tsserver")),
+        // Turns off all frontend-asset processing: no tsserver process, no JS/TS
+        // files or Owl XML templates from manifest asset bundles are parsed or
+        // added to the symbol table, no JS diagnostics/features.
+        FieldSpec::new(DisableJavascript, "disable_javascript", Bool)
+            .default(ConfigValue::bool(false))
+            .triggers_restart(),
     ]
 }
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -371,14 +371,18 @@ impl SyncOdoo {
                 }
             }
         }
-        let tsserver_handle = session.clone_sender_to_main().map(|sender| {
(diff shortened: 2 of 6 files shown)


server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -678,7 +678,7 @@ impl SymbolTable {
         if path_str.ends_with(".py") || path_str.ends_with(".pyi") || FileMgr::is_untitled(&path_str) {
             return Some(session.st_mut().add_new_file(parent, &name, &path_str).into());
         }
-        if path_str.ends_with(".js") {
+        if path_str.ends_with(".js") && !session.sync_odoo.config.are_frontend_assets_disabled() {
```
fda-odoo: I'm less sure about this one. If the user disabled frontend_assets, do we still want to disable custom js entrypoint?

server/src/core/odoo.rs
```diff
@@ -2462,10 +2466,20 @@ impl Odoo {
         //Odoo::update_file_index(session, path,true, false, false);
     }
 
+    /// Whether `extension` (no leading dot) is a source kind the server processes.
+    /// JS/TS are excluded when `disable_frontend_assets` is set.
+    fn is_recognized_extension(session: &SessionInfo, extension: &str) -> bool {
+        match extension {
+            "py" | "xml" | "csv" => true,
```
fda-odoo, line 2570: add .pyi

server/src/core/config/config_key_spec.rs
```diff
@@ -95,6 +96,12 @@ pub(super) fn specs() -> Vec<FieldSpec> {
             .triggers_restart(),
         FieldSpec::new(TsServerCommand, "tsserver_command", Str) //it could maybe be nice to resolve config variables like {workspaceFolder}
             .triggers_restart()
-            .default(ConfigValue::str("tsserver"))
+            .default(ConfigValue::str("tsserver")),
+        // Turns off all frontend-asset processing: no tsserver process, no JS/TS
+        // files or Owl XML templates from manifest asset bundles are parsed or
+        // added to the symbol table, no JS diagnostics/features.
+        FieldSpec::new(DisableFrontendAssets, "disable_frontend_assets", Bool)
```
fda-odoo: I'm not sure that a config with the name "disable frontend" should deactivate tsserver, but only the frontend. I would rather change the name, or let tsserver activated. Let tsserver active, but without project is not really expensive

server/src/core/odoo.rs
mmahrouss: But there was no `.pyi`  check for `update_file_cache` and `update_file_index` do we want to include it here? we do not track updates on `pyi` files anyway, no?

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: Hmmm yes you have a point, we do not need to have the check here, but that means we do not need the check at `update_file_cache` and `update_file_index` too, no?

server/src/core/config/config_key_spec.rs
mmahrouss: Hmmmm okay, so we should scope it to just skip loading assets, but we let tsserver run, and also we process JS/TS files normally ( as external EPs), right?

fda-odoo (approved):

PR #654, opened 2026-07-20
[IMP] standardization of error codes

fda-odoo:
Source is now always "Odoo"
Code is now:
`tsserver_` + code from tsserver
`oxc_` + code from oxc
`OLS**` for python diagnostics


cammarosano (approved):
Nice!

PR #658, opened 2026-07-24
[IMP] complete and validate xml_ids in python strings

andg-odoo:
Hover and go-to-definition already resolved xml_id string literals, but completion and validation did not exist for them. Add an `XML_ID` expected type wired to `env.ref()` (via the `EvalEnvRef `hook), the `groups=` field keyword, and `has_group`/`has_groups` on r`es.users` (`user_has_groups` on earlier versions), with candidates enumerated by prefix from the module dependency graph and filtered to r`es.groups` where the position implies it. `groups=` entries complete per comma-segment and honor the '!' negation prefix from `odoo.tools.set_expression`.

Validate the same positions: `env.ref()` with an unknown local id now raises `OLS05001` and unresolvable `groups=` entries raise `OLS05054`, matching the codes already used for the XML-side checks. Only literal strings are diagnosed; dynamically built values are ignored.

### DIFF
--- a/server/src/core/csv_validation.rs
+++ b/server/src/core/csv_validation.rs
@@ -7,7 +7,7 @@ use tracing::info;
 use crate::{
     Sy, constants::{BuildStatus, BuildSteps, DEBUG_STEPS, DiagnosticSource, MissingDataSource, OYarn}, core::{
         diagnostics::{DiagnosticCode, create_diagnostic}, evaluation_utils::DeepFieldEvalWalker, file_mgr::FileInfo, symbols::{SymbolTable, symbol_keys::{CsvFileKey, ModuleKey}}
-    }, features::csv_ast_utils::CsvFieldIter, oyarn, threads::SessionInfo
+    }, features::csv_ast_utils::{CsvFieldIter, split_csv_xml_ids}, oyarn, threads::SessionInfo
 };
 use std::{cell::RefCell, rc::Rc};
 
@@ -119,11 +119,16 @@ impl CsvValidator {
 
     fn validate_record(&self, session: &mut SessionInfo, csv_module: ModuleKey, headers_is_xml: &[bool], record: &StringRecord, diagnostics: &mut Vec<Diagnostic>, data: &str) {
         let Some(field_iter) = CsvFieldIter::new(record, data) else { return; };
-        for (idx, (start, end, field)) in field_iter.enumerate() {
+        for (idx, (start, _end, field)) in field_iter.enumerate() {
             let Some(&should_be_xml_id) = headers_is_xml.get(idx) else { break;};
-            if should_be_xml_id {
+            if !should_be_xml_id {
+                continue;
+            }
+            // the cell spans its quotes when it carries them, the ids inside it do not
+            let value_start = start + usize::from(data.as_bytes().get(start) == Some(&b'"'));
+            for (xml_id, id_start, id_end) in split_csv_xml_ids(field, value_start) {
                 //check that field_data is a valid xml id
-                let id_split = field.split(".").collect::<Vec<&str>>();
+                let id_split = xml_id.split(".").collect::<Vec<&str>>();
                 let mut module_name = session.st().name(csv_module).as_str();
                 if id_split.len() == 2 {
                     module_name = id_split.first().unwrap();
@@ -134,7 +139,7 @@ impl CsvValidator {
                 let Some(&module_symbol) = session.sync_odoo.modules.get(module_name) else {
                     if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05003, &[]) {
                         diagnostics.push(Diagnostic {
-                            range: Range { start: Position::new(start as u32, 0), end: Position::new(end as u32, 0) },
+                            range: Range { start: Position::new(id_start as u32, 0), end: Position::new(id_end as u32, 0) },
                             ..diagnostic.clone()
                         });
                     }
@@ -147,7 +152,7 @@ impl CsvValidator {
                     session.sync_odoo.get_main_entry().borrow_mut().not_found_data_ids.insert(csv_module.into());
                     if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05001, &[]) {
                         diagnostics.push(Diagnostic {
-                            range: Range { start: Position::new(start as u32, 0), end: Position::new(end as u32, 0) },
+                            range: Range { start: Position::new(id_start as u32, 0), end: Position::new(id_end as u32, 0) },
                             ..diagnostic.clone()
                         });
                     }
--- a/server/src/core/csv_validation.rs
+++ b/server/src/core/csv_validation.rs
@@ -7,7 +7,7 @@ use tracing::info;
 use crate::{
     Sy, constants::{BuildStatus, BuildSteps, DEBUG_STEPS, DiagnosticSource, MissingDataSource, OYarn}, core::{
         diagnostics::{DiagnosticCode, create_diagnostic}, evaluation_utils::DeepFieldEvalWalker, file_mgr::FileInfo, symbols::{SymbolTable, symbol_keys::{CsvFileKey, ModuleKey}}
-    }, features::csv_ast_utils::CsvFieldIter, oyarn, threads::SessionInfo
+    }, features::csv_ast_utils::{CsvFieldIter, split_csv_xml_ids}, oyarn, threads::SessionInfo
 };
 use std::{cell::RefCell, rc::Rc};
 
@@ -119,11 +119,14 @@ impl CsvValidator {
 
     fn validate_record(&self, session: &mut SessionInfo, csv_module: ModuleKey, headers_is_xml: &[bool], record: &StringRecord, diagnostics: &mut Vec<Diagnostic>, data: &str) {
         let Some(field_iter) = CsvFieldIter::new(record, data) else { return; };
-        for (idx, (start, end, field)) in field_iter.enumerate() {
+        for (idx, (start, _end, field)) in field_iter.enumerate() {
             let Some(&should_be_xml_id) = headers_is_xml.get(idx) else { break;};
-            if should_be_xml_id {
+            if !should_be_xml_id {
+                continue;
+            }
+            for (xml_id, id_start, id_end) in split_csv_xml_ids(field, data, start) {
                 //check that field_data is a valid xml id
-                let id_split = field.split(".").collect::<Vec<&str>>();
+                let id_split = xml_id.split(".").collect::<Vec<&str>>();
                 let mut module_name = session.st().name(csv_module).as_str();
                 if id_split.len() == 2 {
                     module_name = id_split.first().unwrap();
@@ -134,7 +137,7 @@ impl CsvValidator {
                 let Some(&module_symbol) = session.sync_odoo.modules.get(module_name) else {
                     if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05003, &[]) {
                         diagnostics.push(Diagnostic {
-                            range: Range { start: Position::new(start as u32, 0), end: Position::new(end as u32, 0) },
+                            range: Range { start: Position::new(id_start as u32, 0), end: Position::new(id_end as u32, 0) },
                             ..diagnostic.clone()
                         });
                     }
@@ -147,7 +150,7 @@ impl CsvValidator {
                     session.sync_odoo.get_main_entry().borrow_mut().not_found_data_ids.insert(csv_module.into());
                     if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05001, &[]) {
                         diagnostics.push(Diagnostic {
-                            range: Range { start: Position::new(start as u32, 0), end: Position::new(end as u32, 0) },
+                            range: Range { start: Position::new(id_start as u32, 0), end: Position::new(id_end as u32, 0) },
                             ..diagnostic.clone()
                         });
                     }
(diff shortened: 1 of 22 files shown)


server/src/core/python_arch_eval_hooks.rs
```diff
@@ -1367,6 +1369,16 @@ impl PythonArchEvalHooks {
...
+        if in_validation && !xml_id.is_empty() && !xml_id.contains('.') {
+            let is_known = ModuleSymbol::get_xml_id(session.st(), module_key, &xml_id)
+                .is_some_and(|xml_ids| xml_ids.iter_valid(session.st()).next().is_some());
+            if !is_known && let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS05001, &[]) {
+                diagnostics.push(Diagnostic {
+                    range: FileMgr::textRange_to_temporary_Range(&xml_id_expr.range()),
+                    ..diagnostic
+                });
+            }
+        }
```
mmahrouss: One thing that is important to check with adding this diagnostic, is checking on odoo+enterprise if we have any false positives. Otherwise it is nice to have of course

andg-odoo: Hey @mmahrouss!

Thanks for the first look, I decided to go and sweep `--parse` across 16.0+ on comm/enterprise and see what this adds versus plain 1.5.1, and yeah sadly it adds too many false positives, I'm not sure how many of them can be directly handled or not, so I'd love yours and the teams opinion on the best way forward. Every version came back with 0 removed and 0 changed diagnostics, so everything below is purely what this PR adds, all of it in `.py`. Here are the stats of new errors generated per branch:

| | 16.0 | 17.0 | 18.0 | 19.0 | master |
|---|---|---|---|---|---|
| as submitted | 461 | 528 | 591 | 340 | 355 |
| after the follow-up commits below | 423 | 484 | 525 | 264 | 260 |

I looked at some of the cases and found parts that can be fixed as bugs in my code and put them into the followup commits:

1. **`groups=fields.NO_ACCESS`**: the sentinel resolves to `'.'`, so every field using it got
   *"Group(s) with id(s) '.' does not exist"*: `mail_message.py`, `ir_attachment.py`, `auth_totp`,
   all the `social_*` modules. **Fixed**

2. Registry-generated ir.model.data: `base.module_*` (`env.ref('base.module_account').demo` is a
   common idiom), `<mod>.field_<model>__<field>`, `<mod>.selection__*`. These only exist after install,
   never in the sources. 35 to 67 hits depending on version. Skipped by shape;
   `module_*` is scoped to `base` so a real record with that prefix elsewhere is still checked. 

   **I've skipped these for now.** For `field_*` and `selection__*` it would mean an xml_id per field (30k+ field definitions on master) plus resolving which module owns each inherited one, so I'd rather not pay that unless we want it. `base.module_*` is much cheaper, one entry per module against the existing `ModuleKey`, and `XmlId` only has one exhaustive match site, so if you'd prefer that one indexed rather than skipped, it's a small change and would give goto-def on modules as a bonus.

3. `groups=` only checked that the xml_id exists, not that it is a `res.groups` record, a typo
   landing on any other record passed silently. `XmlArchBuilder::get_group_ids` already filters on the
   record model, so this just aligns the python side with the XML side. Verified with a probe
   (`groups='base.main_company'` = flagged, `groups='base.group_user'` = clean); zero hits on real
   code across all five versions. **Fixed**

Here is what is left and not handled, as of now master has 260 remaining errors and 185 of them are just in test files.

| category | count | notes |
|---|---|---|
| no static declaration anywhere | 152 | `stock` alone is 146: `stock_location_stock`, `picking_type_in/out` are created by warehouse code at install time |
| declared in a manifest `demo` file | 93 | `base.user_demo`, `product.product_product_4`. `manifest_create.rs` only loads the `data` key |
| declared under the manifest's `other_files` key | 9 | `hr.employee_sj`, `marketing_automation.mail_template_body_welcome_template` are loaded at runtime via `convert.convert_file()`, not at install, yet still always exist at runtime |
| declared in no manifest key at all | 4 | `mail.mail_template_test`, `sms.sms_template_test` test xml loaded by the test itself |
| rows skipped by a pre-existing `OLS05069` | 2 | `crm_iap_mine`'s industry csv: 73 short rows never reach `insert_xml_id` |

All of stock's cases sit in `tests/` files: 100% on 16.0 through 19.0, and 144 of 146 on master. The two exceptions are `stock.stock_reception_report_action`, referenced from `stock/models/stock_picking.py` and `mrp/models/mrp_production.py`, which isn't declared anywhere either. Not checking for proper ref's in the test files doesn't seem like the right solution but I'm not sure what is: it is tempting, since skipping `tests/` cuts 71-86% depending on the version, more than every indexing fix combined.

What makes me think it is mostly test noise: once test files are excluded the remainder is flat at 76 / 70 / 77 / 77 / 75 from 16.0 to master, while the test-file count swings between 180 and 443. The production-code hits are a small stable set of real patterns and nearly all of the version to version variance is test churn.

The large-ish demo bucket is mainly because as of now `manifest_create.rs` only handles `data` files, `demo` and `other_files` buckets are never loaded. This means all demo xml_id documents look like they are incorrect and even though `other_files` aren't run at install time, they still would have goto-def/ref functionality that would be useful. Fixing those would remove ~100 of master's 260. The remaining ~150 are records that genuinely do not exist until install and can't be known statically.

It does its job well though, it's found some records that moved modules a while ago and have been incorrect since 16.0 which is cool, but now we need to know how correct we want to get it.

What do you guys think is the best solution? Indexing `demo` and `other_files` is probably important and would solve a good chunk, but I can't think of a clean way to handle the stock ones, so maybe this should be a warning and not an error in python.

server/src/core/csv_validation.rs
```diff
@@ -119,11 +119,16 @@ impl CsvValidator {
...
         let Some(field_iter) = CsvFieldIter::new(record, data) else { return; };
-        for (idx, (start, end, field)) in field_iter.enumerate() {
+        for (idx, (start, _end, field)) in field_iter.enumerate() {
             let Some(&should_be_xml_id) = headers_is_xml.get(idx) else { break;};
-            if should_be_xml_id {
+            if !should_be_xml_id {
+                continue;
+            }
+            // the cell spans its quotes when it carries them, the ids inside it do not
+            let value_start = start + usize::from(data.as_bytes().get(start) == Some(&b'"'));
```
mmahrouss: I would move the logic of the quotation into `split_csv_xml_ids`

server/src/core/odoo.rs
```diff
@@ -1524,6 +1524,50 @@ impl SyncOdoo {
...
+            if !names_module {
+                continue;
+            }
+            let in_deps = ModuleSymbol::is_in_deps(session.st(), current_module, dir_name);
+            for (local_id, xml_ids) in session.st()[module_key].xml_ids.iter().filter(|(local_id, _)| local_id.starts_with(id_prefix.unwrap_or(""))) {
+                if xml_ids.iter_valid(session.st()).any(|xml_id| match model_filter {
+                    None => true,
+                    Some(model) => matches!(xml_id, XmlId::XmlRecord(record_key) if session.st()[record_key].model.0 == *model),
+                }) {
+                    results.push((module_key, local_id.clone(), in_deps));
```
mmahrouss, line 1557: Wouldn't it be better to only autocomplete the module names if the use is just typing the module, seems like better UX, because let's say they start typing `a` they will have a huge list, maybe they should be able to choose the module, then choose the xml_id? what do you think? that would need renaming the fn, but it is okay

server/src/core/python_arch_eval_hooks.rs
mmahrouss: @andg-odoo 
Sorry for the late reply. 

I think `demo` and `other_files` could be okay to implement. 

I would leave the fixes in your PR, as they really close good gaps in ourcode.

However, I think `OLS05001` should be `Disabled` by default
by changing it to `OLS05001, DiagnosticSetting::Disabled, "Unknown XML ID",` 


Then perhaps we can ask runbot to enable it to catch new bugs, so it improves future code, but enabling it by default might be too noisy in the editor for most people. 

I would need @fda-odoo's opinion too on the state of `OLS5001` and if he agrees or has other suggestions

mmahrouss (commented):
### [FIX] validate each id of a relational csv column
Looks good, we really missed that case, thank you! 
I would add it too in `get_symbols_in_record` like so
```diff
diff --git a/server/src/features/csv_ast_utils.rs b/server/src/features/csv_ast_utils.rs
index 2a77b798..e732a10b 100644
--- a/server/src/features/csv_ast_utils.rs
+++ b/server/src/features/csv_ast_utils.rs
@@ -1,8 +1,12 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+
 use csv::{Reader, StringRecord};
 use lsp_types::Range;
 use ruff_text_size::{TextRange, TextSize};
 
 use crate::core::evaluation_utils::DeepFieldEvalWalker;
+use crate::core::file_mgr::FileInfo;
 use crate::core::symbols::storage::xml::xml_field_symbol::XmlFieldName;
 use crate::core::symbols::symbol_keys::CsvFileKey;
 use crate::features::goto_utils::{GotoSource, GotoSourceType};
@@ -204,7 +208,18 @@ impl CsvAstUtils {
         }
         if headers.contains(&oyarn!("id")) {
             for record in csv_reader.records().filter_map(Result::ok) {
-                    CsvAstUtils::get_symbols_in_record(session, offset, &headers, &record, &mut results, file_symbol, main_symbol.into(), module, content);
+                CsvAstUtils::get_symbols_in_record(
+                    session,
+                    offset,
+                    &headers,
+                    &record,
+                    &mut results,
+                    file_symbol,
+                    main_symbol.into(),
+                    module,
+                    content,
+                    file_info.clone(),
+                );
             }
         }
         results
@@ -220,6 +235,7 @@ impl CsvAstUtils {
         main_symbol: SymbolKey,
         module: Option<ModuleKey>,
         content: &str,
+        file_info: Rc<RefCell<FileInfo>>,
     ) {
         let Some(field_iter) = CsvFieldIter::new(record, content) else { return; };
         //Search for selected field
@@ -263,18 +279,31 @@ impl CsvAstUtils {
                 return;
             };
             // We found the relational model referred to, we can lookup the xml_id now
+            let value_start = start + usize::from(content.as_bytes().get(start) == Some(&b'"'));
+            let Some((xml_id, start, end)) = split_csv_xml_ids(field_data.as_str(), value_start).find(
+                |(_, start, end)| offset >= *start && offset <= *end,
+            ) else {
+                return;
+            };
             results.extend(
                 SyncOdoo::get_xml_ids(
                     session,
                     csv_symbol.into(),
-                    field_data.as_str(),
+                    xml_id,
                     &std::ops::Range::default(), //we don't care about range as it's used only for diagnostic
                     &mut vec![],
                 )
                 .iter_valid(session.st())
                 .map(|xml_id| GotoSource {
                     source: GotoSourceType::SymbolKey(xml_id.into()),
-                    origin_selection_range: None,
+                    origin_selection_range: Some(Range {
+                        start: file_info
+                            .borrow()
+                            .offset_to_position(start as u32, session.sync_odoo.encoding),
+                        end: file_info
+                            .borrow()
+                            .offset_to_position(end as u32, session.sync_odoo.encoding),
+                    }),
                 }),
             );
         }

```

### [FIX] read raise_if_not_found in every spelling
Nice! 

### [FIX] index the demo and other_files manifest keys
Cool!

### [IMP] complete and validate xml_ids in python strings
It fills many gaps, good addition

I Have one comment about completion, as well as the one here in this message.

server/src/core/python_arch_eval_hooks.rs
andg-odoo: If we set it to disabled we would lose the xml and csv side that is currently enabled and working well. This PR only expands to the python side of potential errors. Would it potentially make sense to split it out to two different error messages? Python only then everything else?

Note that the numbers I listed above of errors generated have shrunk dramatically since that comment, here is the "new"ish numbers by adding demo, other_files, and the rest of the fixes in this PR:

| | 16.0 | 17.0 | 18.0 | 19.0 | master |
|---|---|---|---|---|---|
| original PR | 461 | 528 | 591 | 340 | 355 |
| after the first follow-up commits | 423 | 484 | 525 | 264 | 260 |
| **now** | **319** | **406** | **445** | **154** | **154** |
| in `tests/` | 316 | 389 | 427 | 135 | 142 |
| in production code | 3 | 17 | 18 | 19 | 12 |
| distinct ids flagged | 22 | 51 | 47 | 51 | 46 |

It cut out an extra hundred or so records across the different versions which is a good amount, but doesn't account for runtime created records like tax reports and the stock issues (It's still on average between 70%-90% of the False positives coming from stock)

It's interesting that the majority of the errors come from only a few distinct xmlids, it's a very small list that just appears over and over :laugh:

Let me know how we want to approach this, happy with whatever solution you guys decide.

server/src/core/odoo.rs
andg-odoo: Makes sense, right now typing can suggest up to 300k possible xmlids worst case depending on addons paths. Swapping it down to module then record works and `env.ref` already requires a scoped module, you can't do just a base xmlid in it.

However, what are your thoughts on keeping the current flow for `groups=` as there is only 300 records total there and it could be useful without having to have it be two step.

I tried that and it's what I've pushed if you want to take a look, adding `get_xml_id_modules` next to `get_xml_ids_by_prefix` that will be used to populate a list of module names, (auto complete would complete `module.`) then populate the rest of the ids scoped on the module.

mmahrouss: yes I think it is a good idea to keep the current behavior for groups, but change it for xmlids in env.ref
cool:)

PR #659, opened 2026-07-24
[FIX] properly resolve the type of env.user and other properties

andg-odoo:
`self.env.user` and other properties on `env` resolved to nothing: hovering them showed a function, and chains through them (`self.env.user.has_group(...)`, or through a variable) resolved to nothing at all.

Three things were in the way, one per commit: 
- `@functools.cached_property` was not recognized as a property, so Odoo's `Environment.user`/`company`/`companies`/`lang` were plain methods
- `follow_ref` never descended into a property, so reading one as an attribute gave the function instead of the type it returns, for every property and not only Odoo's
- Their annotations are useless anyway (`BaseModel`, `str | None`), so `user`/`company` now resolve their model from the model index at evaluation time, like `self.env["res.users"]` already does.

Unrelated but noticed on the way: the `Environment.registry` file hook no longer does anything from 18.1 on. `registry` became a property there, and a file hook on a function is overwritten by `handle_func_evaluations` from the return annotation, which is why it kept looking fine (`-> Registry` does the job on its own). It is still what types `registry` on 17.0, where it is a plain attribute without annotation, so I left it alone.

### DIFF
--- a/server/src/core/evaluation.rs
+++ b/server/src/core/evaluation.rs
@@ -144,7 +144,7 @@ impl ExprOrIdent<'_> {
  * diagnostics: a vec the hook can fill to add diagnostics
  * file_symbol: if provided, can be used to add dependencies
  */
-type GetSymbolHookCallable = fn (session: &mut SessionInfo, eval: &EvaluationSymbol, context: Option<&Context>, diagnostics: &mut Vec<Diagnostic>, scope: Option<SymbolKey>) -> Option<EvaluationSymbolPtr>;
+pub type GetSymbolHookCallable = fn (session: &mut SessionInfo, eval: &EvaluationSymbol, context: Option<&Context>, diagnostics: &mut Vec<Diagnostic>, scope: Option<SymbolKey>) -> Option<EvaluationSymbolPtr>;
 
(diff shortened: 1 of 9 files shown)


fda-odoo:
Hey @andg-odoo !
I will look closer into this feature, as I wanted this for the next version, and as I'm trying to bring back the "attribute not found".
I checked your pr, and here are my thoughts:
- first commit seems perfect to me. Exactly what I wanted to implement (except maybe the module that is not checked for lazy_property)
- Second seems wrong to me and could be removed. I agree that follow_ref isn't going through properties, but returning the property function is indeed right. When we'll want to evaluate the property, we will take the return value of the function, and here is the difference for properties.
- Third seems right, but it's sad that the hooks doesn't follow the syntax of other hooks of the file.
I will take it from here and check that everything is ok, as well as rebase it on the other pending branches (that are modifying follow_ref too)
Thank you for the pr ;)

fda-odoo:
New PR is here: https://github.com/odoo/odoo-ls/pull/700 :)

andg-odoo:
Awesome! Glad to hear it is progressing!

PR #660, opened 2026-07-28
Alpha unpack tuple fda

fda-odoo:


server/tests/test_unpack.rs
```diff
@@ -0,0 +1,100 @@
...
+    let (mut odoo, config) = setup::setup::setup_server(false);
+    let mut session = setup::setup::create_init_session(&mut odoo, config);
+    let path = env::current_dir().unwrap().join("tests/data/python/expressions/for_loop_unpack.py").sanitize();
+    setup::setup::prepare_custom_entry_point(&mut session, path.as_str());
+    assert!(session.sync_odoo.entry_point_mgr.borrow().custom_entry_points.len() == 1);
+    let st = &session.sync_odoo.symbol_table;
+
+    let int_type = session.sync_odoo.get_symbol("", (&["builtins"], &["int"]), u32::MAX)[0];
+    let str_type = session.sync_odoo.get_symbol("", (&["builtins"], &["str"]), u32::MAX)[0];
+    let x = session.sync_odoo.get_symbol(path.as_str(), (&[], &["x"]), 26);
```
cammarosano: These offsets are not really helpful to see which line is being tested.
I would suggest to refer to those by line, and use `FileInfo::position_to_offset` to convert them..
Or simply (like you said), just use different variable names and use u32::MAX.

cammarosano: Suggestion:
I think `x` should be the result of `get_symbol` followed by `first().unwrap()`, so you don't need these `x[0]` everywhere.

server/src/core/python_utils.rs
```diff
@@ -87,6 +85,58 @@ fn _link_tuples(targets: Vec<Expr>, values: Vec<Expr>) -> Vec<Assign> {
...
+The assignations stop at 20 assigns. If more values are found, last assign will be a "unknown" value (None) and stop at 21
+Ex: for a in [1, 2, 3], return [[("a", None, 1), ("a", None, 2), ("a", None, 3)]]
+for a, b in [(1, 2), (3, 4)], return [[("a", None, 1), ("b", None, 2)], [("a", None, 3), ("b", None, 4)]]
+*/
+pub fn unpack_iter_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<&Expr>) -> Vec<Vec<Assign>> {
+    let mut res = Vec::new();
+    if let Some(value) = value {
+        match value {
+            Expr::List(list) => {
+                for elt in list.elts.iter().take(20) {
```
cammarosano: you should probably assing `20` to a `MAX_ASSIGNS` constant and refer to it, instead of magic `20` in all these places.

server/src/core/python_utils.rs
```diff
@@ -18,7 +17,6 @@ pub struct Assign {
 fn _link_tuples(targets: Vec<Expr>, values: Vec<Expr>) -> Vec<Assign> {
```
cammarosano:
```suggestion
fn _link_tuples(targets: &[Expr], values: &[Expr]) -> Vec<Assign> {
```
As we don't need to own these vectors here, we could take them as slices, so the caller doesn't need to clone vectors.

server/src/core/python_utils.rs
```diff
@@ -145,58 +195,60 @@ pub fn unpack_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<
...
+                if let Some(value) = value {
+                    if value.is_tuple_expr() {
+                        res.append(&mut _link_tuples(expr.elts.clone(), value.clone().tuple_expr().unwrap().elts.clone()));
+                        return res;
+                    } else if value.is_list_expr() {
+                        res.append(&mut _link_tuples(expr.elts.clone(), value.clone().list_expr().unwrap().elts.clone()));
+                        return res;
+                    } else if value.is_set_expr() {
+                        res.append(&mut _link_tuples(expr.elts.clone(), value.clone().set_expr().unwrap().elts.clone()));
+                        return res;
```
cammarosano: see comment in `_link_tuples`. We shouldn't have to clone these `elts` vectors, as we just need to read them.

server/src/core/python_utils.rs
```diff
@@ -145,58 +195,60 @@ pub fn unpack_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<
...
+                    }
```
cammarosano, line 229: same as above (we probably don't need clones of `elts`)

server/src/core/python_utils.rs
```diff
@@ -145,58 +195,60 @@ pub fn unpack_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<
...
                 // the type of the value, we return the value with an index
-                if value.is_none() {
-                    //debug!("Invalid stmt: can't annotate a tuple");
-                    continue;
+                if let Some(value) = value {
+                    if value.is_tuple_expr() {
+                        res.append(&mut _link_tuples(expr.elts.clone(), value.clone().tuple_expr().unwrap().elts.clone()));
+                        return res;
+                    } else if value.is_list_expr() {
+                        res.append(&mut _link_tuples(expr.elts.clone(), value.clone().list_expr().unwrap().elts.clone()));
```
cammarosano: instead of this `is_tuple_expr()` with later `tuple_expr().unwrap()` pattern,
why not `match value { Expr::Tuple => {...}, Expr::List => {...} }`  ?
This applies to code in `_link_tuples` too (not really changed by this PR).

server/src/core/python_utils.rs
```diff
@@ -87,6 +85,58 @@ fn _link_tuples(targets: Vec<Expr>, values: Vec<Expr>) -> Vec<Assign> {
...
 
+/*
+Given an Expr, generate Assigns for each variable in the target, but as if targets are searching for an iterable value (as in a for-loop).
+The assignations stop at 20 assigns. If more values are found, last assign will be a "unknown" value (None) and stop at 21
+Ex: for a in [1, 2, 3], return [[("a", None, 1), ("a", None, 2), ("a", None, 3)]]
+for a, b in [(1, 2), (3, 4)], return [[("a", None, 1), ("b", None, 2)], [("a", None, 3), ("b", None, 4)]]
+*/
+pub fn unpack_iter_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<&Expr>) -> Vec<Vec<Assign>> {
+    let mut res = Vec::new();
+    if let Some(value) = value {
```
cammarosano: you could avoid the code repetition, considering `elts` has always the same type:

```rust
let Some(value) = value else { return unpack_assign(targets, annotation, None) };
let elts = match value {
    Expr::List(list) => list.elts,
    Expr::Tuple(tuple) => tuple.elts,
    Expr::Set(set) => set.elts,
    _ => { return unpack_assign(targets, annotation, None) }
};
let mut res = Vec::new();
for elt in elts.iter().take(20) {
    let unpacked = unpack_assign(targets, annotation, Some(elt));
    res.push(unpacked);
}
// this for-loop above could be a map btw, if you want:
//let mut res = elts.iter().take(20).map(|elt| unpack_assign(targets, annotation, Some(elt));
if elts.len() > 20 {
    let unpacked = unpack_assign(targets, annotation, None);
    res.push(unpacked);
}

```

server/src/core/python_arch_eval.rs
```diff
@@ -804,35 +806,86 @@ impl PythonArchEval {
...
+                    continue;
+                };
+                let variable = session.st().get_positioned_symbol(*self.sym_stack.last().unwrap(), &assign_target.id, &assign_target.range());
+                variables.push(variable);
+            }
+        }
+        //iterate through all different assignation to associate the different evaluations
+        for (index, variable) in variables.iter().enumerate() {
+            let Some(variable) = variable else {continue};
+            let variable = variable.unwrap_variable_key();
```
cammarosano, line 826: Unless you want it to panic for a non-variable symbol, this is simpler with:
```suggestion
            let &Some(SymbolKey::Variable(variable)) = variable else {continue};
```

server/src/core/python_arch_eval.rs
```diff
@@ -804,35 +806,86 @@ impl PythonArchEval {
...
+                            value,
+                            *self.sym_stack.last().unwrap(),
+                            &for_stmt.target.range().start(), false, &mut deps);
+                        session.st_mut().insert_dependencies(self.file, &deps, self.current_step);
+                        self.diagnostics.extend(diags);
+                        for eval_iter in eval_iter_node.iter() {
+                            let eval_symbol = eval_iter.symbol.get_symbol(session, None, &mut vec![], None);
+                            let symbol_eval = SymbolTable::follow_ref(&eval_symbol, session, None, false, false, None, None);
+                            let Some(symbol_type) = symbol_eval[0].upgrade_weak(session.st()) else {continue};
+                            let symbol = eval_iter.symbol.get_symbol_as_weak(session, Some(&Context::from_iter([(ContextKey::ParentFor, ContextValue::SYMBOL(symbol_type.into()))])), &mut vec![], None);
```
cammarosano: That's a long line.
Maybe build the Context before it?

cammarosano (changes requested):
Mostly style and DRY suggestions. You can just ignore them if you don't want to apply them.
The ones I would ask for changes are
- the unnecessary Vector clones (as it could impact our build time),
- the offsets in the test (hard to see what's being tested without a dedicated extension...)

server/src/core/python_arch_eval.rs
fda-odoo: yes, I want to panic here. Something else than a variable is unexpected, so I would want to have a crash if it happen

cammarosano (approved):

PR #661, opened 2026-07-28
Address clippy stuff

cammarosano:
Small refactors, to get rid of some `[allow(clippy::*]`

### DIFF
--- a/server/src/core/pre_parser.rs
+++ b/server/src/core/pre_parser.rs
@@ -64,6 +64,12 @@ const SKIPPED_DIRS: &[&str] = &[
     "i18n", "views", "data", "security", "description", "doc", "docs",
 ];
 
+/// (owner module path, local url)
+/// local url contains globs, not expanded yet
+type AssetEntry = (String, String);
+/// The result of expanding an AssetEntry: the list of files it matches.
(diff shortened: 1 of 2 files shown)


server/src/core/pre_parser.rs
```diff
@@ -64,6 +64,12 @@ const SKIPPED_DIRS: &[&str] = &[
     "i18n", "views", "data", "security", "description", "doc", "docs",
 ];
 
+/// (owner module path, local url)
+/// local url cointains globs, not expanded yet
```
fda-odoo: contains

server/src/core/pre_parser.rs
```diff
@@ -64,6 +64,12 @@ const SKIPPED_DIRS: &[&str] = &[
...
+type AssetEntry = (String, String); 
+/// The result of expanding a AssetEntry: the list of files it matches.
```
fda-odoo: an AssetEntry

fda-odoo (approved):
great !

server/src/core/pre_parser.rs
cammarosano: :grimacing: Thanks!

mmahrouss (approved):

PR #663, opened 2026-07-28
[TESTS] semantic tokens python

cammarosano:


mmahrouss (approved):

fda-odoo (approved):
Nice! I'll wait for the fix of 'website' before uncomment the test here and merge

cammarosano:
> Nice! I'll wait for the fix of 'website' before uncomment the test here and merge

I rewrote the `todo` so that it fails when the fix lands - and doesn't let us forget to toggle the `todo` to a `check`.
What do you think?

Do you still prefer that I remove the last commit (non-passing test case)?

fda-odoo:
@cammarosano No, it's ok for me like that! we'll see the test when fixing it, so it's fine to me.
Can you rebase on alpha?

PR #664, opened 2026-07-29
[FIX] assume text_document can be void

fda-odoo:
text_document on file_info_ast can be empty (if file stream is invalid, ie not valid UTF-8), but some places in the code were assuming that text_document is always present

### DIFF
--- a/server/src/core/csv_validation.rs
+++ b/server/src/core/csv_validation.rs
@@ -41,7 +41,10 @@ impl CsvValidator {
             session.st_mut().set_build_status(csv_symbol.into(), BuildSteps::VALIDATION, BuildStatus::INVALID);
             return;
         };
-        let data = file_info.borrow().file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
+        let Some(data) = file_info.borrow().file_info_ast.borrow().text_document.as_ref().map(|td| td.contents().to_string()) else {
+            // File can be invalid (not valid UTF-8 and so text_document is empty)
+            return;
+        };
         let model_name_pb = Path::new(&path);
         let model_name = model_name_pb.file_stem().unwrap().to_str().unwrap();
         let csv_module = session.st().find_module(csv_symbol);
--- a/server/src/core/symbols/manifest_arch_eval.rs
+++ b/server/src/core/symbols/manifest_arch_eval.rs
@@ -57,11 +57,10 @@ impl ModuleSymbol {
                 let csv_sym = session.st_mut().add_new_csv_file(symbol_key, &file_name, &path_string);
                 Self::on_data_file_load(session.st(), csv_sym.into());
                 session.st_mut().add_dependency(symbol_key.into(), csv_sym.into(), BuildSteps::ARCH_EVAL, BuildSteps::ARCH);
-                if file_info.file_info_ast.borrow().text_document.as_ref().is_none() {
-                    //TODO do we want to add a diagnostic here?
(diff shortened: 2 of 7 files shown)


server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -57,8 +57,7 @@ impl ModuleSymbol {
                 let csv_sym = session.st_mut().add_new_csv_file(symbol_key, &file_name, &path_string);
                 Self::on_data_file_load(session.st(), csv_sym.into());
                 session.st_mut().add_dependency(symbol_key.into(), csv_sym.into(), BuildSteps::ARCH_EVAL, BuildSteps::ARCH);
-                if file_info.file_info_ast.borrow().text_document.as_ref().is_none() {
-                    //TODO do we want to add a diagnostic here?
+                if file_info.file_info_ast.borrow().text_document.as_ref().is_none() { // File can be invalid (not valid UTF-8 and so text_document is empty)
                     continue;
                 }
                 let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
```
mmahrouss: In this case we can do let else, what do you think?

server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -155,15 +154,17 @@ impl ModuleSymbol {
...
             ModuleSymbol::load_xml_arch(session, xml_sym, &mut file_info, true);
         }
     }
 
     fn load_xml_arch(session: &mut SessionInfo, xml_sym: XmlFileKey, file_info: &mut FileInfo, web_asset: bool) {
+        if file_info.file_info_ast.borrow().text_document.is_none() { // File can be invalid (not valid UTF-8 and so text_document is empty)
+            return;
+        }
         //That's a little bit crappy, but the SYNTAX step of XML files are done here, as lifetime of roXMLTree are not flexible enough to be separated from the Arch building
         let data = file_info.file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
```
mmahrouss: also here

server/src/core/csv_validation.rs
```diff
@@ -41,6 +41,11 @@ impl CsvValidator {
             session.st_mut().set_build_status(csv_symbol.into(), BuildSteps::VALIDATION, BuildStatus::INVALID);
             return;
         };
+        //text_document is not guaranteed to be present, as the file can be invalid (not valid UTF-8 for example)
+        if file_info.borrow().file_info_ast.borrow().text_document.is_none() {
+            session.st_mut().set_build_status(csv_symbol.into(), BuildSteps::VALIDATION, BuildStatus::INVALID);
+            return;
+        }
         let data = file_info.borrow().file_info_ast.borrow().text_document.as_ref().unwrap().contents().to_string();
```
mmahrouss: +1

mmahrouss: and in other places in the PR

mmahrouss (approved):
A good opportunity to eliminate some unwraps, otherwise LGTM

server/src/core/symbols/manifest_arch_eval.rs
cammarosano: Suggestion: this "borrow to check if none, them borrow again to unwrap" pattern could be replaced by:
```rust
let Some(data) = file_info.file_info_ast.borrow().text_document.as_ref().map(|td| td.contents().to_string()) else {
   // File can be invalid (not valid UTF-8 and so text_document is empty)
    continue;
};
```

cammarosano (commented):
Nice, fast fix!

PR #667, opened 2026-07-29
[FIX] Fix goto location when going to python package.

fda-odoo:
Add missing __init__.py component to the uri

### DIFF
--- a/server/src/features/goto_utils.rs
+++ b/server/src/features/goto_utils.rs
@@ -260,6 +260,12 @@ impl GotoUtils {
                             Range::default()
                         )
                     },
+                    SymType::PACKAGE(PackageType::PYTHON_PACKAGE) => {
+                        (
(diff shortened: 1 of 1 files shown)


mmahrouss (approved):

server/src/features/goto_utils.rs
```diff
@@ -260,6 +260,12 @@ impl GotoUtils {
                             Range::default()
                         )
                     },
+                    SymType::PACKAGE(PackageType::PYTHON_PACKAGE) => {
+                        (
+                            Some(Path::new(&session.st().path(symbol_key.as_source_file_key().unwrap())).join("__init__.py").sanitize()),
```
cammarosano: just call `st().file_path` instead.

If the match is done directly on the key (not `.typ()`) it's just:
```rust
SymbolKey::PythonPackage(p) => (
    session.st().file_path(p.into()),
    Range::default()
)
```

PR #670, opened 2026-07-29
Fix config errors

mmahrouss:
1. Use default config if there are workspace folders or config file path
2. Log invalid names in config file to the client


fda-odoo (approved):
Great! It could be nice to put the "odoo button" in red if toml is invalid too. Actually I think it's showing a notification, but if it disappear, nothing is left to show it

mmahrouss:
> Great! It could be nice to put the "odoo button" in red if toml is invalid too. Actually I think it's showing a notification, but if it disappear, nothing is left to show it

Yes good idea, I will add that in a third commit

fda-odoo (approved):

PR #671, opened 2026-07-29
[FIX] Do not set dependency on custom JavaScript files

mmahrouss:
Js Validator calls `get_main_entry` which panics if we do not have odoo path set

### DIFF
--- a/server/src/core/js_validator.rs
+++ b/server/src/core/js_validator.rs
@@ -38,20 +38,22 @@ impl JsValidator {
         let template_refs = file_info_ast.ast.as_js_ast().js_template_refs.clone();
         drop(file_info_ast);
 
-        for template_ref in template_refs.iter() {
-            if !XmlAstUtils::ensure_js_template_validity(session, &template_ref.t_name) {
-                session.st_mut()[self.js_symbol].not_found_data_ids.insert(MissingDataSource::TEMPLATE(Sy!(template_ref.t_name.clone())), BuildSteps::VALIDATION);
-                session.sync_odoo.get_main_entry().borrow_mut().not_found_data_ids.insert(self.js_symbol.into());
-                if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS06000, &[]) {
-                    diagnostics.push(Diagnostic {
-                        range: Range { start: Position::new(template_ref.range.start().to_u32(), 0), end: Position::new(template_ref.range.end().to_u32(), 0) },
-                        ..diagnostic
-                    });
-                }
-            } else if let Some(imps) = session.sync_odoo.js_templates.get_mut(&template_ref.t_name) {
-                for template_imp in imps.iter_valid(&session.sync_odoo.symbol_table) {
-                    let xml_file = session.st().get_file(template_imp.into()).expect("Template should be in a file");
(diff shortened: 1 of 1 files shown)


server/src/core/js_validator.rs
```diff
@@ -41,7 +41,9 @@ impl JsValidator {
         for template_ref in template_refs.iter() {
             if !XmlAstUtils::ensure_js_template_validity(session, &template_ref.t_name) {
                 session.st_mut()[self.js_symbol].not_found_data_ids.insert(MissingDataSource::TEMPLATE(Sy!(template_ref.t_name.clone())), BuildSteps::VALIDATION);
-                session.sync_odoo.get_main_entry().borrow_mut().not_found_data_ids.insert(self.js_symbol.into());
+                if session.sync_odoo.has_main_entry {
+                    session.sync_odoo.get_main_entry().borrow_mut().not_found_data_ids.insert(self.js_symbol.into());
```
fda-odoo: I think the issue here was not to check if has_main_entry is true, but it should have inserted not_found_data_ids to the right entry_point (maybe a custom one) instead of the main (issue maybe present elsewhere in the code btw)

cammarosano:
```rust
let entry = session.st().get_entry(self.js_symbol);
entry.borrow_mut().not_found_data_ids.insert(self.js_symbol.into());
```
Like fda said, the `get_main_entry()`  use is seen in other places too, but for things we believe that would always be found in the main entry point  (xml and csv) .

But then we got a crash report that suggests otherwise: https://iap-services.odoo.com/odoo/action-970/2245

```
2026-07-29T07:30:00.798309Z  INFO ThreadId(07) odoo_ls_server::core::entry_point: Adding new custom entry point: /home/odoo/Desktop/odoo/addons/l10n_in_edi/views/account_journal_dashboard_views.xml
2026-07-29T07:30:00.798362Z  INFO ThreadId(07) odoo_ls_server::core::entry_point: Adding new custom entry point: /home/odoo/Desktop/odoo/addons/l10n_jp/data/account_report_jct_accumulation_form.xml
2026-07-29T07:30:00.798393Z  INFO ThreadId(07) odoo_ls_server::core::entry_point: Adding new custom entry point: /home/odoo/Desktop/odoo/addons/l10n_jp/data/account_report_jct_accumulation_schedule_1_3.xml
2026-07-29T07:30:00.798426Z  INFO ThreadId(07) odoo_ls_server::core::entry_point: Adding new custom entry point: /home/odoo/Desktop/odoo/addons/l10n_jp/data/account_report_jct_accumulation_schedule_2_3.xml
```

So maybe we should consider adapting all usages of `get_main_entry` + `not_found_data_ids.insert` to call `get_entry()` instead.

cammarosano (changes requested):

server/src/core/js_validator.rs
fda-odoo: I think we should check first why these custom entry points are created for xml/csv files. I'm not sure it's on purpose and that's more a bug

mmahrouss: As discussed IRL, this block is intended to check for missing dependencies between QWeb templates and JS files, which should only be present in Main EPs, I  changed the condition to surround the whole for-loop, to see if the file belongs to the main entry point or not

fda-odoo (approved):

cammarosano (approved):

PR #675, opened 2026-07-30
BuildScheduler refactoring

fda-odoo:


fda-odoo:
It's ready for review, even if I want to drive more tests by comparing it with the alpha branch. But tests are passing, and manual manipulation seems right too. If more changes would come they should be small

fda-odoo:
Still not sure about the target version also, but probably still 1.5.2 as it is fixing various bugs (mostly race conditions, so annoying ones).

server/src/core/import_resolver.rs
```diff
@@ -409,7 +409,9 @@ fn resolve_new_symbol(session: &mut SessionInfo, parent: SymbolKey, imported_nam
             || session.sync_odoo.is_file_cs(&full_path.join("__init__.pyi").sanitize_cow())
         ) {
             if let Some(symbol) = SymbolTable::create_from_path(session, &full_path, parent, false) {
-                SyncOdoo::build_now(session, symbol, BuildSteps::ARCH);
+                if let Some(buildable) = symbol.as_buildable_symbol_key() {
+                    BuildScheduler::build_now(session, buildable, BuildSteps::ARCH);
+                }
```
mmahrouss, line 414: It seems that `build_now` only does something for `PythonBuildableSymbolKey` not for the general `BuildableSymbolKey`, so perhaps we could change the signature of `build_now` to accept `PythonBuildableSymbolKey` instead? 

Or perhaps there is an issue at `build_now` ?

fda-odoo: I kept it generic, but indeed the only usage seems to be for python files. I can restrict build_now to that indeed!

fda-odoo: I was wrong. build_now will trigger build_now_dependencies that call back build_now, and a python file could depend on an xml file

server/src/core/symbols/symbol_keys.rs
```diff
@@ -410,3 +417,87 @@ pub enum JsFileParent {
...
+        match *self {
+            SymbolKey::Function(k) => Some(BuildableSymbolKey::Function(k)),
+            SymbolKey::File(k) => Some(BuildableSymbolKey::File(k)),
+            SymbolKey::Module(k) => Some(BuildableSymbolKey::Module(k)),
+            SymbolKey::PythonPackage(k) => Some(BuildableSymbolKey::PythonPackage(k)),
+            SymbolKey::XmlFile(k) => Some(BuildableSymbolKey::XmlFile(k)),
+            SymbolKey::CsvFile(k) => Some(BuildableSymbolKey::CsvFile(k)),
+            SymbolKey::JsFile(k) => Some(BuildableSymbolKey::JsFile(k)),
+            _ => None,
+        }
```
cammarosano: You could do this instead:
```suggestion
       BuildableSymbolKey::try_from(*self).ok()
```

server/src/core/symbols/symbol_keys.rs
```diff
@@ -410,3 +417,87 @@ pub enum JsFileParent {
...
+        match *self {
+            SymbolKey::Function(k) => BuildableSymbolKey::Function(k),
+            SymbolKey::File(k) => BuildableSymbolKey::File(k),
+            SymbolKey::Module(k) => BuildableSymbolKey::Module(k),
+            SymbolKey::PythonPackage(k) => BuildableSymbolKey::PythonPackage(k),
+            SymbolKey::XmlFile(k) => BuildableSymbolKey::XmlFile(k),
+            SymbolKey::CsvFile(k) => BuildableSymbolKey::CsvFile(k),
+            SymbolKey::JsFile(k) => BuildableSymbolKey::JsFile(k),
+            _ => panic!("Not a buildable symbol key"),
+        }
```
cammarosano: Same as above (+ expect). You could do this:
```suggestion
        BuildableSymbolKey::try_from(*self).expect("Not a buildable symbol key")
```

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -456,53 +456,71 @@ impl SymbolTable {
...
-            SymbolKey::XmlDelete(_) => panic!(),
-            SymbolKey::CsvFile(k) => self[k].build_status(step),
-            SymbolKey::JsFile(j) => self[j].build_status(step),
+            BuildableSymbolKey::PythonPackage(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::Module(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::File(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::Function(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::XmlFile(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::CsvFile(k) => self[k].get_build_status(step),
+            BuildableSymbolKey::JsFile(j) => self[j].get_build_status(step),
```
cammarosano, line 468: beautiful! ❤

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1049,10 +1063,12 @@ impl SymbolTable {
                 }
                 for (index, sym) in build_queue {
                     if index + 1 == BuildSteps::ARCH_EVAL as usize {
-                        session.sync_odoo.add_to_rebuild_arch_eval(sym);
+                        session.st_mut().reset_build_status(sym.into(), BuildSteps::ARCH_EVAL, BuildStatus::PENDING);
+                        BuildScheduler::queue(session, sym);
                     } else if index + 1 == BuildSteps::VALIDATION as usize {
                         SymbolTable::invalidate_sub_functions(session, sym);
-                        session.sync_odoo.add_to_validations(sym);
+                        session.st_mut().reset_build_status(sym.into(), BuildSteps::ARCH_EVAL, BuildStatus::PENDING);
```
cammarosano: as discussed IRL, take a look at this (should be BuildSteps::VALIDATION ?).
Also consider making `reset_build_status` more strict, preventing it from setting an invalid step (like `set_build_status`).

server/src/core/python_arch_builder.rs
```diff
@@ -45,22 +47,16 @@ pub struct PythonArchBuilder {
...
-            return None; // nothing to extract
-        }
-        if symbol_table.build_status(symbol, BuildSteps::ARCH) != BuildStatus::PENDING {
-            return None;
-        }
-        let file = symbol_table.get_file(symbol).unwrap();
-        let file_mode = symbol == file;
+    pub fn new(symbol_table: &SymbolTable, entry_point: Rc<RefCell<EntryPoint>>, symbol: PythonBuildableSymbolKey) -> Option<Self> {
+        let file = symbol_table.get_file(symbol.into()).unwrap();
+        let file_mode = SymbolKey::from(symbol) == SymbolKey::from(file);
```
cammarosano, line 52: couldn't we just check if symbol matches PythonBuildableSymbolKey::Function instead?

server/src/core/python_arch_builder.rs
```diff
@@ -179,7 +180,6 @@ impl PythonArchBuilder {
         if self.file_mode {
             PythonArchBuilderHooks::on_file_done(session, self.file);
         }
-        session.st_mut().set_build_status(self.sym_stack[0], BuildSteps::ARCH, BuildStatus::DONE);
```
cammarosano, line 182: as discussed IRL, take a look at this. If we don't hit the if nor the else-if above, we leave the status as IN_PROGRESS forever...

server/tests/setup/setup.rs
```diff
@@ -30,7 +31,7 @@ pub fn setup_server(with_odoo: bool) -> (SyncOdoo, ConfigEntry) {
     let subscriber = FmtSubscriber::builder()
         .with_thread_ids(true)
         .with_file(false)
-        .with_max_level(LevelFilter::INFO)
+        .with_max_level(LevelFilter::TRACE)
```
cammarosano: this was for debug purposes only, I suppose?

cammarosano:
Nice! It seems we have a performance improvement!
```
commit                user (s)           sys (s)     total CPU (s)          wall (s)       peak RSS (MB)
──────────  ────────────────  ────────────────  ────────────────  ────────────────  ──────────────────
4fb3e7b9       20.952 ± 0.117    2.696 ± 0.047   23.648 ± 0.129   18.704 ± 0.078       2826.1 ± 1.8
59961b86       20.106 ± 0.029    2.698 ± 0.028   22.804 ± 0.029   17.854 ± 0.030       2823.7 ± 0.9

Δ (after − before):
  user (s):              -0.846  ( -4.04%)   z=7.0
  sys (s):               +0.002  ( +0.07%)   z=0.0
  total CPU (s):         -0.844  ( -3.57%)   z=6.4
  wall (s):              -0.850  ( -4.54%)   z=10.1
  peak RSS (MB):           -2.3  ( -0.08%)   z=1.1

```

cammarosano (changes requested):
Cool! A nice improvement in organization and also better typing! Love it!
Just please check a few risks I noted as review comments...

server/src/core/python_arch_builder.rs
fda-odoo: Yes, it would work too. I just kept the test as it was before. I like it this way because it resistant with types update

server/tests/setup/setup.rs
fda-odoo: Indeed :) we don't see messages generated by DEBUG_STG constants in tests. I reverted it

fda-odoo:
@cammarosano I adressed your comments in a new commit. Let me know what you think about the reset_build_status changes. If we don't want that I think we would have to do multiple BuildSteps enums

cammarosano:
> @cammarosano I adressed your comments in a new commit. Let me know what you think about the reset_build_status changes. If we don't want that I think we would have to do multiple BuildSteps enums

I think it's good like this. Now about a related matter: the code duplication.
@fda-odoo @mmahrouss  Here's an idea/suggestion to avoid code duplication in the Buildable trait: https://github.com/odoo/odoo-ls/commit/ed734d94d1f81d4446f6da2d314695970cb2bcd8
What do you think?

mmahrouss:
> > @cammarosano I adressed your comments in a new commit. Let me know what you think about the reset_build_status changes. If we don't want that I think we would have to do multiple BuildSteps enums
> 
> I think it's good like this. Now about a related matter: the code duplication. @fda-odoo @mmahrouss Here's an idea/suggestion to avoid code duplication in the Buildable trait: [ed734d9](https://github.com/odoo/odoo-ls/commit/ed734d94d1f81d4446f6da2d314695970cb2bcd8) What do you think?

Looks nice I like it:)

cammarosano:
https://github.com/odoo/odoo-ls/pull/698

mmahrouss (approved):

cammarosano (approved):

PR #678, opened 2026-08-04
[FIX] Check that AST is built during JS  Validation

mmahrouss:
For Custom EPs for JS files created through `handle_did_create` their ASTs are not built yet, so before we attempt to validate them, we build the AST if needed

This draws a parallel to the check in python_arch_builder. 
This is made to fix crash report  2260

### DIFF
--- a/server/src/core/file_mgr.rs
+++ b/server/src/core/file_mgr.rs
@@ -436,7 +436,7 @@ impl FileInfo {
             return false;
         }
         self.file_info_ast.borrow_mut().text_hash = new_hash;
-        self.diagnostics.clear();
+        self.clear_diagnostics();
         self._build_ast(session, is_external);
         true
     }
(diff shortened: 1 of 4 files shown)


cammarosano (approved):

fda-odoo:
I don't really see why AST is not ready. Usually file_info updates should always be done directly and never deferred.
From what I see handle_did_create is indeed not updating the file_info. I think it is because create is only an information. DidOpen is the real update of the file with versioned content.
Maybe we shouldn't create a custom entry point on didCreate, but only on DidOpen? Or update the fileinfo on didCreate too (but that means directly from disk, not from lsp)

mmahrouss:
> I don't really see why AST is not ready. Usually file_info updates should always be done directly and never deferred. From what I see handle_did_create is indeed not updating the file_info. I think it is because create is only an information. DidOpen is the real update of the file with versioned content. Maybe we shouldn't create a custom entry point on didCreate, but only on DidOpen? Or update the fileinfo on didCreate too (but that means directly from disk, not from lsp)

Hmmm I think yes it is not a bad idea to actually not create custom entry points from did_create, and only in did open, maybe I should do that, if so I can close this PR and open a new one with that edit

mmahrouss:
@fda-odoo Could you please check the last 2 commits, see if that is a change we would like to have?

fda-odoo (approved):

PR #681, opened 2026-08-06
[REF] lifecycle: parent enum

cammarosano:
No more Holder traits, no more compile-time checks (other than exhaustive pattern-matching).

Follow-up of this PR: ban symbol key replacement when inserting to a hashmap with key (name, path) collision.

The actual LOC diff is smaller than it seems: tests have been added.

Note: This PR supersedes https://github.com/odoo/odoo-ls/pull/676. So in case this one gets merged first, we can just bring the tests from it.


fda-odoo (approved):

mmahrouss (approved):
Looks cool!

PR #682, opened 2026-08-06
[ADD] README to debug_tools

cammarosano:
And toggle ignoreFailures to false in launch.json, so that if the pretty-print debug script doesn't load, we get an error.


mmahrouss (approved):

fda-odoo (approved):

fda-odoo:
Note to myself before merging: add a link to this file in wiki

PR #683, opened 2026-08-07
[FIX] Recover lost FileInfo

mmahrouss:
In a rare scenario with multiple custom entrypoints that interact with each other, one FileInfo can be deleted because it is external, while the main file is still queued, but in ARCH_EVAL, so it does not recover and lead to a panic, and also leaves a wrong in_progress status on the symbol.

The test also triggers this scenario, before this fix, you get a panic

This might be fixed by #675 as a side effect, but still the test might be valuable to keep

fixes crash report 2272

### DIFF
--- a/server/src/core/csv_validation.rs
+++ b/server/src/core/csv_validation.rs
@@ -6,7 +6,7 @@ use tracing::info;
 
 use crate::{
     Sy, constants::{BuildStatus, BuildSteps, DEBUG_STEPS, DiagnosticSource, MissingDataSource, OYarn}, core::{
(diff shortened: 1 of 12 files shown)


fda-odoo (changes requested):
This idea is already present for python validation step with get_file_info_for_validation. And I guess it's the reason you only update arch and arch_eval here.
Maybe we could merge this two functions?

mmahrouss:
> This idea is already present for python validation step with get_file_info_for_validation. And I guess it's the reason you only update arch and arch_eval here. Maybe we could merge this two functions?

Hmm yes good idea!

fda-odoo (approved):

PR #684, opened 2026-08-11
[IMP] Add Warning diagnostic to detect indirectly inherited methods/f…

mmahrouss:
…ields

Note, this adds ~450 diagnostics on Odoo+enterprise, I skimmed through them and they seem okay

I am not sure what version we should target here? perhaps 1.5.2?

### DIFF
--- a/server/src/core/diagnostic_codes_list.rs
+++ b/server/src/core/diagnostic_codes_list.rs
@@ -194,6 +194,13 @@ OLS03026, DiagnosticSetting::Error, "Invalid value for operator: 'access'; The v
 */
 OLS03027, DiagnosticSetting::Error, "Invalid field for operator: 'access'; The field should be a model id or a Many2One field to a model",
 /**
+* You are accessing a method or field that only exists because of a model extension (_inherit)
+* declared in a module that is not in the dependencies of the current module. This works only
+* because that module currently happens to be installed alongside yours ("indirect inheritance")
+* and can break if it's ever made optional or uninstalled. Add it to your manifest's depends.
+*/
+OLS03028, DiagnosticSetting::Warning, "'{0}' is only defined through indirect inheritance from module '{1}', which are not declared as a dependency",
+/**
 * Form is no longer available on odoo.tests.common, thus it should not be imported from there.
 */
 OLS03301, DiagnosticSetting::Warning, "Deprecation Warning: Since 17.0: odoo.tests.common.Form is deprecated, use odoo.tests.Form",
--- a/server/src/core/evaluation.rs
+++ b/server/src/core/evaluation.rs
@@ -1120,7 +1120,20 @@ impl Evaluation {
                                     }
                                     evals.push(eval);
                                 });
-                            }
+                            } else if is_in_validation && module.is_some() {
+                                    // Retry unfiltered: if it resolves at all, it's only reachable through
+                                    // a module outside our dependencies ("indirect inheritance").
+                                    let (indirect_syms, _) = SymbolTable::get_member_symbol(session, base_loc, &expr.attr, None, false, false, false, false, is_super);
+                                    if let Some(indirect_module) = indirect_syms.first().and_then(|&sym| session.st().find_module(sym)) {
+                                        let dir_name = session.st()[indirect_module].dir_name.clone();
+                                        if let Some(diagnostic) = create_diagnostic(session, DiagnosticCode::OLS03028, &[&expr.attr, dir_name.as_str()]) {
+                                            diagnostics.push(Diagnostic {
+                                                range: FileMgr::textRange_to_temporary_Range(&expr.range()),
+                                                ..diagnostic
+                                            });
+                                        }
+                                    }
+                                }
                         }
                     }
                 }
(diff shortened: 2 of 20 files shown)


server/src/core/evaluation.rs
```diff
@@ -1096,7 +1096,19 @@ impl Evaluation {
                                     }
                                     evals.push(eval);
                                 });
-                            }
+                            } else if is_in_validation
+                                && let Some(from_module) = module {
+                                    let indirect_modules = SymbolTable::get_indirect_inheritance_modules(session, base_loc, &expr.attr, from_module);
```
fda-odoo: It seems too complex to me. As you call this method only when you have no results by calling get_member_symbol with a from_module, couldn't you just call it again with from_module=None? If you have results you would know that it's outside dependencies. Here you are building the list of 'not in dependencies' modules for each attribute, it seems too much.

mmahrouss: The issue is in `get_member_symbol_helper` if `from_module=None` we look for the module of the current symbol, so that will not work. Yet perhaps we can still simplify the search though, I will look for a cheaper solution

mmahrouss: I updated `get_member_symbol_helper` and the whole check

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1300,7 +1280,7 @@ impl SymbolTable {
         if !matches!(base_attr, SymbolKey::Class(_)) {
             return res;
         }
-        //TODO shouldn't we set the from_module in the call to get_member_symbol?
+
```
fda-odoo, line 1283: so it was useless to set it in the end?

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -2031,13 +2011,22 @@ impl SymbolTable {
         Self::member_symbol_hook(session, target, name, &mut diagnostics);
         let mod_sym = session.st().get_module_symbol(target, name);
         if let Some(mod_sym) = mod_sym
-            && !only_fields {
+            && !only_fields
+        {
+            let in_scope = match (mod_sym, from_module) {
```
fda-odoo: I would call it 'in_deps' or something like that, as 'scope' wording is more about range in a file and sections

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: Yep, because we wanted to get "__get__" on the current class, not go into odoo inheritence. to update `__get__` anyway on a field, you would need to define a new field and use Python inheritence

server/src/core/symbols/symbol_table_impl.rs
mmahrouss: Agreed and done:D

fda-odoo (approved):

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -2098,18 +2087,34 @@ impl SymbolTable {
         if model_data.is_some() && !prevent_comodel {
             let model = session.sync_odoo.models.get(&model_data.as_ref().unwrap().name).cloned();
             if let Some(model) = model {
-                let mut from_module = from_module;
-                if from_module.is_none() {
-                    from_module = session.st().find_module(target);
+                // from_module: None means no dependency filtering (return everything)
```
cammarosano, line 2090: maybe add this to the function's docstring as well?

server/src/core/diagnostic_codes_list.rs
```diff
@@ -194,6 +194,13 @@ OLS03026, DiagnosticSetting::Error, "Invalid value for operator: 'access'; The v
 */
 OLS03027, DiagnosticSetting::Error, "Invalid field for operator: 'access'; The field should be a model id or a Many2One field to a model",
 /**
+* You are accessing a method or field that only exists because of a model extension (_inherit)
+* declared in a module that is not in the dependencies of the current module. This works only
+* because that module currently happens to be installed alongside yours ("indirect inheritance")
+* and can break if it's ever made optional or uninstalled. Add it to your manifest's depends.
+*/
+OLS03028, DiagnosticSetting::Warning, "'{0}' is only defined through indirect inheritance from module(s) '{1}', which are not declared as a dependency",
```
cammarosano: why plural in `module(s)`?

server/src/core/evaluation.rs
```diff
@@ -1096,7 +1096,20 @@ impl Evaluation {
                                     }
                                     evals.push(eval);
                                 });
-                            }
+                            } else if is_in_validation && module.is_some() {
+                                    // Retry unfiltered: if it resolves at all, it's only reachable through
+                                    // a module outside our dependencies ("indirect inheritance").
+                                    let (indirect_syms, _) = SymbolTable::get_member_symbol(session, base_loc, &expr.attr, None, false, false, false, false, is_super);
```
cammarosano, line 1126: this has a perf hit...
about 4% increase in first build time

cammarosano (commented):
I don't fully understand what's going on here, but LGTM. Just beware of the perf hit in case you were not aware...

server/src/core/diagnostic_codes_list.rs
mmahrouss: That is from the previous iteration, now it only catches one, I will update it

server/src/core/evaluation.rs
mmahrouss: Okay perhaps we can just pass from_module to None, then checking the results after the call and see if that is faster. I will try that and see

PR #685, opened 2026-08-12
[IMP] allow user to disable semantic tokens

fda-odoo:
We already got feedbacks that some people doesn't like that, so we don't want to force it and make a way to disable it for people who want it


mmahrouss (approved):

cammarosano (approved):

cammarosano:
In `Server::initialize` we should not include a `semantic_tokens_provider` when the config disables it.
The way it is, we advertise that we provide semantic tokens, and then just return None for each request.

PR #687, opened 2026-08-13
Move Models. FileInfos and EntryPoints to SlotMaps

mmahrouss:


mmahrouss:
As discussed with the team, we can better manage the localization of the `SlotMap`s for `Model`s, `FileInfo`s, and `EntryPoint`s.

- Convert the `HashMap` `SyncOdoo.models` into its own type `ModelMgr`
- Move the `SlotMap`s for all the three we migrated into their managers
- Update the lifecycle, getters and setters

mmahrouss:
Performance of the latest commit: 

image

server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1002,19 +997,6 @@ impl SymbolTable {
...
-            SourceFileKey::PythonPackage(p) => &mut self[p].model_dependencies,
-            SourceFileKey::File(f) => &mut self[f].model_dependencies,
-            SourceFileKey::XmlFile(x) => &mut self[x].model_dependencies,
-            SourceFileKey::CsvFile(c) => &mut self[c].model_dependencies,
-            SourceFileKey::JsFile(j) => &mut self[j].model_dependencies,
-        };
-        model_dependencies.insert(model.clone());
-        model.borrow_mut().add_dependent(target);
-    }
-
```
cammarosano: worth mentioning on the commit message that you removed the `model_dependencies` field of all these symbols (dead code).
Otherwise this comes as a surprise to the reviewer, who expects all diff outside of model.rs to be just adaptation to the new API of ModelMgr.

server/src/core/model.rs
```diff
@@ -19,6 +19,77 @@ use crate::threads::SessionInfo;
...
+        self.models_index.get(name).map(|key| &self[*key])
+    }
+
+}
+
+impl std::ops::Index<ModelKey> for ModelMgr {
+    type Output = Model;
+    // This is safe for two reasons:
+    // 1. Models are never removed
+    // 2. We call it usually after `get_model_key` but that is not very trustworthy
```
cammarosano: what do you mean by " that is not very trustworthy"?

cammarosano (commented):
I have reviewed only the first commit for now.
I'll add a second review later.
So far: amazing!

PR #688, opened 2026-08-13
[FIX] Avoid panic on missing type annotation

mmahrouss:
Let's say we `x, y: int` or `x, y: ` or `[x, y] : str`, those are all invalid, but they were getting None for value and annotation, which makes `handle_assigns` panic with "either value or annotation should exists". Now we pass the annotation, but if we have the invalid syntax, or a non-single target, we only create an evaluation if we have a value, so an assignment operator `=`

See crash report 2273

### DIFF
--- a/server/src/core/python_utils.rs
+++ b/server/src/core/python_utils.rs
@@ -197,7 +197,8 @@ pub fn unpack_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<
                 }
                 for (index, target) in expr.elts.iter().enumerate() {
                     match target {
-                        Expr::Name(tar) => {
+                        // Annotation and value can be None when evaluation for loop var in arch build and in document symbol
+                        Expr::Name(tar) if annotation.is_none() || value.is_some() => {
                             res.push(Assign {
                                 target: AssignTargetType::Name(tar.clone()),
                                 annotation: None,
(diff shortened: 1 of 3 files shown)


server/src/core/python_utils.rs
```diff
@@ -197,10 +197,11 @@ pub fn unpack_assign(targets: &[Expr], annotation: Option<&Expr>, value: Option<
                 }
                 for (index, target) in expr.elts.iter().enumerate() {
                     match target {
-                        Expr::Name(tar) => {
+                        // Annotation and value can be None when evaluation for loop var in arch build
+                        Expr::Name(tar) if annotation.is_none() || value.is_some() => {
                             res.push(Assign {
                                 target: AssignTargetType::Name(tar.clone()),
-                                annotation: None,
+                                annotation: annotation.cloned(),
```
fda-odoo: I don't see in which case passing the annotation here is useful?
The only useful one would be `Tuple[int, int]` we do not support right now.
I think the if three lines above is the fix?

fda-odoo: it could even be wrong if the code by after try to read the xth element of 'int'

mmahrouss: Well indeed yes, it passes the annotation, which do not currently use. 
I added it for correctness, but it does not change the behavior. 

Also, anyway, "Only single target (not tuple) can be annotated" so actually `x, y: int` is invalid syntax, even if we pass it, the user will still get a diagnostic.

Do you think we should revert it back to None? as for me, both options are okay, perhaps Keeping it None is a good idea since the PR is just a fix, and it will just annotate an invalid syntax.

After thinking about it, I think reverting it back to None is the better choice, I updated it now

fda-odoo (approved):

PR #689, opened 2026-08-13
[FIX] Fix panic on class arg for `super()`

mmahrouss:
If the first argument of `super` is `Any`, `None` or `Unbound` `upgrade_weak` can return None, which we were unwraping, causing a panic.

This PR, looks through all the refs hoping to find one valid symbol, it also moves the logic into helper functions to have slightly more readability.

Tests were added that would panic before the fix
Example crash report: 2291

### DIFF
--- a/server/src/core/evaluation.rs
+++ b/server/src/core/evaluation.rs
@@ -252,6 +252,34 @@ fn record_evaluation_hit(session: &mut SessionInfo, parent: SymbolKey, range: Te
     true
 }
 
+/// Resolve and filter super class args
+/// Only returns evaluations that are classes, and returns their instance values
+fn resolve_super_class_args(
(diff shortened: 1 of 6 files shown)


fda-odoo (approved):
LGTM, but I'm wondering if we should maybe take all possible values of the argument instead of only the first valid hit? Maybe the first valid hit doesn't contain the method or attribute we are searching for? What do you think?

mmahrouss:
> LGTM, but I'm wondering if we should maybe take all possible values of the argument instead of only the first valid hit? Maybe the first valid hit doesn't contain the method or attribute we are searching for? What do you think?

yes you are right, I thought about it too, but kept it as it was, but since you suggested, I updated it too now

fda-odoo (approved):

PR #691, opened 2026-08-14
[FIX] handle files being deleted by FS while opened

mmahrouss:
Before that could easily trigger a panic when doing a git operation for example. A deleted file will have its file info deleted, then upon doing any edit, it would panic.

This change prevents deletion if file is opened, but keeps track of that state, if the file gets closed, then the deletion happens at that point.

### DIFF
--- a/server/src/core/python_validator.rs
+++ b/server/src/core/python_validator.rs
@@ -155,9 +155,12 @@ impl PythonValidator {
                 if !session.st().is_external(symbol) {
                     return;
                 }
-                let file = symbol.as_source_file_key().unwrap();
-                let file_path = session.sync_odoo.symbol_table.file_path(file).to_string();
-                FileMgr::delete_file_path(session, &file_path);
+                let file_info = self.file_info.as_ref().unwrap();
+                if !file_info.borrow().opened { // Never delete opened files
+                    let file = symbol.as_source_file_key().unwrap();
+                    let file_path = session.sync_odoo.symbol_table.file_path(file).to_string();
+                    FileMgr::delete_file_path(session, &file_path);
+                }
             } else {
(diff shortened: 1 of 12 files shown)


server/src/core/python_validator.rs
```diff
@@ -169,13 +169,22 @@ impl PythonValidator {
         if matches!(symbol.typ(), SymType::FILE | SymType::PACKAGE(_)) {
             if !session.st().in_workspace(symbol) {
                 if !session.st().is_external(symbol) {
+                    self.file_info.as_ref().unwrap().borrow_mut().pending_delete = false;
```
fda-odoo: why do you need a manual set to false here?

fda-odoo (commented):
Let me know what you think about that, but I'm wondering if we couldn't get rid of the pending_delete field, and simply check everytime we want to close a file if the file is still present or not on the disk, instead of trying to keep an internal pending state that can be corrupted.
DidChangeWatchedFile for deletion or even creation are not guaranteed, so we could lose these notification, and still have pending_delete with an invalid value.
So I would:
- On delete notification -> do not delete if opened
- On close notification -> Delete if file not on disk anymore, checked at this time
- On creation notification -> no need to reset pending_delete to false (that you weren't doing btw)

server/src/core/python_validator.rs
mmahrouss: It is for a file has been revaluated, then we keep it, so we get this if the file was recreated (or from other paths) 

But I do think this logic is not sound after thinking about it again. 

I am rethinking all the scenarios with also the LSP limitations, since we are not supposed to read from disk when the file is opened? but maybe we can check for existance. 

I will ping you when it is ready for review again

fda-odoo (approved):

PR #692, opened 2026-08-14
Alpha hook tests fda

fda-odoo:
Add tests for hooks. Some tests are disabled depending on the odoo version if the hook doesn't apply on it


cammarosano (approved):
❤️‍🔥

mmahrouss (approved):

PR #693, opened 2026-08-17
[FIX] Handle deleted Addons

mmahrouss:
When a directory that is an addon entry is deleted, the symbol gets dropped, but the entrypoint stays. In that case `search_symbols_to_rebuild` calls get_symbol on the entrpoint, then unwraps the result, which then panics.

This PR, adds the mechanism to drop, and then also recover addon entries upon deletion and creation.

See crash report 2271

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -115,6 +115,45 @@ impl TypeshedWeakReferences {
     }
 }
 
+/// Method names for the server's custom "$Odoo/..." LSP notifications, kept in one place so
+/// call sites can't typo a raw string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum OdooNotification {
+    JsLsStatus,
+    InvalidPythonPath,
+    DiagnosticConfig,
+    RestartNeeded,
+    LoadingStatusUpdate,
+    SetConfiguration,
+    SetPid,
+}
+
+impl OdooNotification {
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::JsLsStatus => "$Odoo/jsLsStatus",
+            Self::InvalidPythonPath => "$Odoo/invalid_python_path",
+            Self::DiagnosticConfig => "$Odoo/diagnostic_config",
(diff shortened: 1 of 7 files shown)


fda-odoo (commented):
This PR makes me think about a more general issue: what happen on the module load order when we add a new module/restore addons folders,....
We only build the load module order at startup, not with additional modules.
The actual easy fix would be to reload everything when it happens...

server/src/core/odoo.rs
```diff
@@ -2029,16 +2068,27 @@ impl Odoo {
...
-                    break;
+            let mut addon_entry = ep_mgr.borrow().addons_entry_points.iter()
+                .find(|entry| entry.borrow().path == parent_path_str)
+                .cloned();
+            // path may have been deleted and recreated since; restore it if still configured
+            if addon_entry.is_none() && session.sync_odoo.config.addons_paths().contains(&parent_path_str) {
+                addon_entry = EntryPointMgr::restore_addon_entry(session, &parent_path_str);
+                if addon_entry.is_some() {
+                // Requst restart to get a clean reload
+                // Also queue it later, in case the client does not support `RestartNeeded
```
fda-odoo: Comment is not clear, you are not queuing a restart ;)

mmahrouss: Updated!

server/src/threads.rs
```diff
@@ -55,9 +55,18 @@ impl <'a> SessionInfo<'a> {
         self.sender.clone()
     }
 
+    /// Clone of the receiving end of the same channel `send_notification` et al. write to.
+    /// In production this is used for request/response correlation; in tests (where sender
+    /// and receiver are the same loopback channel) it lets a test drain and inspect whatever
+    /// the session sent out.
+    pub fn clone_receiver(&self) -> Receiver<Message> {
```
fda-odoo: Do not expose the receiver outside of the session, especially if this is only for tests. I would prefer that you create drain_restart_notifications in here instead, with maybe a name as `test_drain_restart_notifications`

mmahrouss: Done!

fda-odoo (approved):

PR #700, opened 2026-08-24
Fix env.user and company / cached_property

fda-odoo:


server/src/core/symbols/symbol_table_impl.rs
```diff
@@ -1360,6 +1372,19 @@ impl SymbolTable {
         res
     }
 
+    /* /!\ we want to keep instance = True if the evaluation we are following was set to True! */
+    fn propagate_instance(next_refs: Vec<EvaluationSymbolPtr>, instance: Option<bool>) -> Vec<EvaluationSymbolPtr> {
+        if !instance.is_some_and(|instance| instance) {
```
mmahrouss: I think the not and then `is_some_and` is a bit hard to read
```suggestion
        // Do nothing if instance is None or False
        if instance.is_none_or(|instance| !instance) {
```

cammarosano (approved):
Great!

PR #705, opened 2026-08-26
[REF] avoid String allocations with oyarn macro

cammarosano:
`oyarn!("{}", name)` avoids malloc for names <= 15 bytes, while `OYarn::from(name.to_string())` always allocates a String.

### DIFF
--- a/server/src/core/symbols/storage/variable_symbol.rs
+++ b/server/src/core/symbols/storage/variable_symbol.rs
@@ -10,6 +10,7 @@ use crate::{
             symbol_keys::{ModelSymbolKey, ModuleKey, SymbolKey, VariableKey},
         },
     },
+    oyarn,
     threads::SessionInfo,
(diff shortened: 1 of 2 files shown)


server/src/core/symbols/storage/variable_symbol.rs
```diff
@@ -31,7 +32,7 @@ impl VariableSymbol {
 
     pub fn new(name: &str, parent: FileContentParent, range: TextRange, is_external: bool) -> Self {
         Self {
-            name: name.to_string().into(),
+            name: oyarn!("{}", name),
```
cammarosano, line 35: this is already the case for other symbols, just `VariableSymbol` was different (I thought they were equivalent when I wrote it)

mmahrouss (approved):

fda-odoo (approved):

PR #706, opened 2026-08-26
[IMP] goto feature now take the header range of functions and classes.

fda-odoo:
It allows us to fallback on gotoreference if the cursor is in the range of the element, but not the body.

### DIFF
--- a/server/src/features/goto_utils.rs
+++ b/server/src/features/goto_utils.rs
@@ -270,8 +270,12 @@ impl GotoUtils {
                         if let Some(file) = session.st().get_file(*symbol_key) {
                             let path = session.st().path(file).to_string();
                             let range = if session.st().has_range(*symbol_key) {
-                                let range = session.st().range(*symbol_key);
-                                session.sync_odoo.get_file_mgr().borrow().text_range_to_range(session, &path, range)
+                                let file_mgr = session.sync_odoo.get_file_mgr();
+                                let file_info = file_mgr.borrow().get_file_info(&path);
+                                let header_range = session.st().header_range(*symbol_key, file_info);
+                                session.sync_odoo.get_file_mgr().borrow().text_range_to_range(session, &path, header_range.unwrap_or(
+                                    session.st().range(*symbol_key)
+                                ))
                             } else {
                                 Range::default()
                             };
--- a/server/src/core/file_mgr.rs
+++ b/server/src/core/file_mgr.rs
@@ -27,7 +27,7 @@ use std::rc::Rc;
 use std::cell::RefCell;
 use crate::S;
 use crate::constants::*;
-use ruff_text_size::{Ranged, TextRange};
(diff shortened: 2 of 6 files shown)


server/src/features/goto_utils.rs
```diff
@@ -270,7 +272,16 @@ impl GotoUtils {
                         if let Some(file) = session.st().get_file(*symbol_key) {
                             let path = session.st().path(file).to_string();
                             let range = if session.st().has_range(*symbol_key) {
-                                let range = session.st().range(*symbol_key);
+                                let mut range = session.st().range(*symbol_key);
+                                if session.st().has_body_range(*symbol_key) {
```
mmahrouss: `has_body_range` and `if let Some(..) = (..).body_range(..)` are redudndant

We could just use body_range and if let some since we always read and check at the same time, we could totally drop has_body_range

server/src/features/goto_utils.rs
```diff
@@ -304,4 +315,29 @@ impl GotoUtils {
...
+        let file_mgr = session.sync_odoo.get_file_mgr();
+        let file_info = file_mgr.borrow().get_file_info(path);
+        if let Some(file_info) = file_info {
+            if file_info.borrow().file_info_ast.borrow().text_document.is_none() {
+                file_info.borrow_mut().prepare_ast(session);
+            }
+            file_info.borrow().file_info_ast.borrow().text_document.as_ref()?;
+            let file_info_ref = file_info.borrow();
+            let ast_ref = file_info_ref.file_info_ast.borrow();
+            let td = ast_ref.text_document.as_ref()?;
```
mmahrouss: This is redudndant from line 328, we can read once here

server/src/features/goto_utils.rs
```diff
@@ -270,7 +272,16 @@ impl GotoUtils {
                         if let Some(file) = session.st().get_file(*symbol_key) {
                             let path = session.st().path(file).to_string();
                             let range = if session.st().has_range(*symbol_key) {
-                                let range = session.st().range(*symbol_key);
+                                let mut range = session.st().range(*symbol_key);
+                                if session.st().has_body_range(*symbol_key) {
+                                    let body_range = session.st().body_range(*symbol_key);
+                                    if let Some(br) = body_range && br.start() > range.start() {
+                                        let end = GotoUtils::prev_line_end(session, &path, br.start())
+                                            .filter(|&end| end > range.start())
```
mmahrouss: We could use `Option::and` I think is clearer with Options than filter

mmahrouss (changes requested):

mmahrouss (approved):
Cool!

fda-odoo:
@mmahrouss Can you review it again? I just fixed the test and moved some code in better places

mmahrouss (approved):

cammarosano (approved):

PR #707, opened 2026-08-27
[FIX] Only add magic fields to main class.

mmahrouss:
This agrees with the framework logic, see `MetaModel`

This is to fix duplicated completions and other features for `create_uid` and the rest

### DIFF
--- a/server/src/core/python_odoo_builder.rs
+++ b/server/src/core/python_odoo_builder.rs
@@ -300,6 +300,13 @@ impl PythonOdooBuilder {
 
     fn _add_magic_fields(&mut self, session: &mut SessionInfo) {
         let symbol = self.symbol;
+        {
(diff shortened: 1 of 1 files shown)


fda-odoo (approved):

server/src/core/python_odoo_builder.rs
```diff
@@ -300,6 +300,13 @@ impl PythonOdooBuilder {
 
     fn _add_magic_fields(&mut self, session: &mut SessionInfo) {
         let symbol = self.symbol;
+        {
+            // Magic fields and log_access fields are only added to the original model definition.
+            let model = session.st_mut()[symbol]._model.as_ref().unwrap();
```
cammarosano: I expect you to expect. 😝

cammarosano (approved):

fda-odoo (approved):

PR #708, opened 2026-08-27
[FIX] Filter duplicate string syms in completion

mmahrouss:
This is also another fix to filter duplicates in completion, in preparation of features for something like : `country_code = fields.Char(related='company_id.country_id.code', depends=['company_id.<trigger completion>']`

### DIFF
--- a/server/src/features/completion.rs
+++ b/server/src/features/completion.rs
@@ -1086,12 +1086,12 @@ fn add_nested_field_names(
                 false,
             );
             for (symbol_name, symbols) in all_symbols {
-                //we could use symbol_name to remove duplicated names, but it would hide functions vs variables
                 if symbol_name.starts_with(name) {
                     let mut found_one = false;
-                    for final_sym in symbols.iter() {
-                        if specific_field_type.is_none() || SymbolTable::is_specific_field(session, *final_sym, &["Many2one", "One2many", "Many2many", specific_field_type.as_ref().unwrap().as_str()]){
-                            items.push(build_completion_item_from_symbol(session, vec![*final_sym], &symbol_name, Context::default()));
+                    // Remove duplicates, except those of different type, where it show a user that a field and a function have a name collision.
+                    for final_sym in symbols.into_iter().unique_by(|&sym| sym.typ()) {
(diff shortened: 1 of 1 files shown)


server/src/features/completion.rs
```diff
@@ -1086,12 +1086,12 @@ fn add_nested_field_names(
                 false,
             );
             for (symbol_name, symbols) in all_symbols {
-                //we could use symbol_name to remove duplicated names, but it would hide functions vs variables
                 if symbol_name.starts_with(name) {
                     let mut found_one = false;
-                    for final_sym in symbols.iter() {
-                        if specific_field_type.is_none() || SymbolTable::is_specific_field(session, *final_sym, &["Many2one", "One2many", "Many2many", specific_field_type.as_ref().unwrap().as_str()]){
-                            items.push(build_completion_item_from_symbol(session, vec![*final_sym], &symbol_name, Context::default()));
+                    // Remove duplicates, except those of the same type, where it show a user that a field and a function have a name collision.
```
fda-odoo: you mean 'except those of a different type'?

server/src/features/completion.rs
```diff
@@ -1086,12 +1086,12 @@ fn add_nested_field_names(
...
+                    for final_sym in symbols.into_iter().unique_by(|&sym| sym.typ()) {
```
fda-odoo, line 1092: if you filter the uniqueness only by type, you are losing a lot of different symbols no? Or I don't get it here.
For a completion of `my_vari`, if you have Variable `my_variable` and the Variable `my_variation`, it will not return the second one, right?

mmahrouss: It is okay because `all_members` returns symbols grouped by name

server/src/features/completion.rs
mmahrouss: yes! oops

cammarosano (commented):
👍

server/src/features/completion.rs
fda-odoo: ok right! indeed !

fda-odoo (approved):

fda-odoo (approved):

PR #710, opened 2026-08-27
[FIX] Filter magic fields on string results

mmahrouss:
It does the same logic for go to utils on normal expressions


cammarosano (approved):
nice refactor

fda-odoo (approved):

Issue #679, opened 2026-08-05
workspace/symbol returns duplicate results (Nx per symbol) from iterating all entry points sharing the same root


chuonglm2012:
**Title:** `workspace/symbol` returns duplicate results (N× per symbol) due to iterating all entry points sharing the same root

**Labels:** bug

---

## Setup

- odoo-ls version: 1.5.1
- Config: `odoo_path = sources/`, `addons_paths = [overlay/community, overlay/enterprise, overlay/custom]`, plus `workspaceFolders = [extra-addons]` sent in `initialize`

## Describe the bug

`workspace/symbol` returns each logical symbol exactly **5 times** — identical `name`, `kind`, `containerName`, and `location.uri`. For symbols with multiple distinct definitions (e.g. `_compute_qc_result` has 5 implementations across different models), the response contains 25 entries (5 distinct × 5 duplicates each).

`textDocument/documentSymbol` (document outline) returns the canonical single copy — the duplication is workspace-level only.

## Root cause

In `workspace_symbols.rs`, `get_workspace_symbols` iterates **all** entry points:

```rust
for entry in ep_mgr.borrow().iter_all() {
    if entry.borrow().typ == EntryPointType::BUILTIN
        || entry.borrow().typ == EntryPointType::PUBLIC {
        continue;
    }
    WorkspaceSymbolFeature::browse_symbol(
        session, &entry.borrow().root, &query, ...
    );
}
```

The issue is that `iter_all()` returns 5 separate entry points that all share the **same root symbol**:

| # | Type  | Path               | Origin                                    |
|---|-------|--------------------|-------------------------------------------|
| 1 | MAIN  | `sources/`         | `set_main_entry` (odoo_path)              |
| 2 | ADDON | `sources/addons/`  | Auto-discovered `odoo_path + "/addons"` (odoo.rs ~L505) |
| 3 | ADDON | `overlay/community`| `addons_paths[0]` (odoo.rs ~L518)         |
| 4 | ADDON | `overlay/enterprise`| `addons_paths[1]`                        |
| 5 | ADDON | `overlay/custom`   | `addons_paths[2]`                         |

All ADDON entry points set `entry.root = related.root` (entry_point.rs ~L160), meaning they share the MAIN entry point's root. `browse_symbol` then performs a full tree traversal for each entry point, yielding identical results N times.

## Possible fixes

[...]

PR #716, opened 2026-08-31
[FIX] Fix duplicates in workspace symbols

mmahrouss:
Fixes #679

### DIFF
--- a/server/src/features/workspace_symbols.rs
+++ b/server/src/features/workspace_symbols.rs
@@ -2,7 +2,7 @@ use lsp_server::{ErrorCode, ResponseError};
 use lsp_types::{Location, SymbolKind, WorkspaceLocation, WorkspaceSymbol, WorkspaceSymbolResponse};
 use ruff_text_size::{TextRange, TextSize};
 
-use crate::{S, constants::SymType, core::{entry_point::EntryPointType, file_mgr::{Ast, FileMgr}, symbols::{storage::{SymbolTable, XmlDataParent}, symbol_keys::{SourceFileKey, SymbolKey, XmlTemplateKey}}}, threads::SessionInfo, utils::string_fuzzy_contains};
+use crate::{S, constants::SymType, core::{entry_point::EntryPointType, file_mgr::{Ast, FileMgr}, symbols::{storage::{SymbolTable, XmlDataParent}, symbol_keys::{SourceFileKey, SymbolKey, XmlTemplateKey}}}, threads::SessionInfo, utils::{HashSet, string_fuzzy_contains}};
 
 /// The prefix OWL template names are offered under, mirroring the `xmlid.` one already used for
 /// XML ids. It keeps a template distinguishable from the component class of the same name, and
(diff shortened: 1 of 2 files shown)


server/src/features/workspace_symbols.rs
```diff
@@ -35,10 +35,16 @@ impl WorkspaceSymbolFeature {
...
                     }
                 }
+        let mut visited_roots = std::collections::HashSet::new();
         for entry in ep_mgr.borrow().iter_all() {
             if entry.borrow().typ == EntryPointType::BUILTIN || entry.borrow().typ == EntryPointType::PUBLIC { //We don't want to search in builtins
                 continue;
             }
+            // Several entry points (e.g. MAIN and its ADDON entry points) can share the same
+            // root symbol; only traverse each root once to avoid duplicate results.
+            if !visited_roots.insert(entry.borrow().root) {
```
fda-odoo, line 45: that's interesting. Maybe it could be worth it to check if any other `for entry in iter_all` could have this filter too?

fda-odoo (approved):

server/src/features/workspace_symbols.rs
mmahrouss: but only addons share the root entry point with the main entrypoint, right? custom entry points can share paths with main, but with different roots, and they should not have overlapping symbols, no?

fda-odoo: Not here, I mean there is maybe other places in the code where this accumulator filter could be useful.
For example, get_symbol in odoo.rs

mmahrouss: Yes it is worth looking at, I will look it into it today

mmahrouss: `get_symbol` does not apply, because it exits early, and even if the root is the same, it can have a different prefix. 
I tried to hash the root and the prefix, and this is  the benchmark: 

```bash

commit                user (s)           sys (s)     total CPU (s)          wall (s)       peak RSS (MB)
──────────  ────────────────  ────────────────  ────────────────  ────────────────  ──────────────────
6b0194f9       19.016 ± 0.133    3.362 ± 0.075   22.378 ± 0.125   15.638 ± 0.139       1734.9 ± 1.2
f3fdafd9       19.264 ± 0.426    3.408 ± 0.080   22.672 ± 0.503   15.744 ± 0.109       1735.7 ± 3.6

Δ (after − before):
  user (s):              +0.248  ( +1.30%)   z=0.6
  sys (s):               +0.046  ( +1.37%)   z=0.4
  total CPU (s):         +0.294  ( +1.31%)   z=0.6
  wall (s):              +0.106  ( +0.68%)   z=0.6
  peak RSS (MB):           +0.9  ( +0.05%)   z=0.2

```

So it is the same or slower at best, and it cannot give us duplicates, I will look at the other methods and report in the next commit

mmahrouss: Apparently all the other methods in odoo.rs already break early, and check prefix. so I think we cannot apply the check somewhere else for now

fda-odoo: Ok great! it is good to know !

fda-odoo (approved):

PR #717, opened 2026-09-01
[REF] Use ASTKind to avoid cloning AST

mmahrouss:
This is probably minor as it only concerns features, but it was an easy win


cammarosano (approved):
Nice!

fda-odoo (approved):
Nice! I like it !

PR #722, opened 2026-09-03
[FIX] JS import completions

cammarosano:
** Don't be scared: many commits are either tests or just renaming. **

Fixes suggestions for completion in
- import statements: `import { somethin`
- elsewhere, with auto-imports: `let a = somethin` -> completion suggests importing from somewhere

2 kinds of fix:
- missing suggestions 
- invalid suggestions:
  - invalid path for odoo loader
  - module not in dependencies


mmahrouss (approved):
I could not follow the whole logic but looks good as far as I can understand

fda-odoo (approved):
Cool !

PR #723, opened 2026-09-04
[REF] Ast::Pending

cammarosano:
Probably better to wait for https://github.com/odoo/odoo-ls/pull/717 to land then rebase.

-------------------
Adds a "Pending" variant to FileMgr::Ast.

Before:
FileInfo initialized with a dummy PythonAst.
So a file_info_ast of PythonAst could mean 2 things: not yet done (but not necessarily related to a .py file), or an actual Python Ast. Which is rather confusing and error-prone.

After:
- Ast::PythonAst can only mean a Python ast - and done (not pending).
- Option around indexed_module removed. If it's a Ast::PythonAst, it contains an indexed_module.
- The only way to represent a pending ast is... Ast::Pending (!)
- is_built is simply: not Pending.
- No more default initialization of PythonAst and JsAst (fields set to empty vectors or None): these are only constructed with the actual parsed data.

Potential bug avoided:
`get_references` was called without confirmation that the Ast was really built after calling prepare_ast. It ast build failed, it would follow the Python arm and panic on file_info_ast.get_stmts().unwrap(). Now the Pending variant is handled (meaning not built) and this is avoided.

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -1532,10 +1532,9 @@ impl Odoo {
                 }
                 let ast_type = file_info.borrow().file_info_ast.borrow().ast.clone();
                 match ast_type {
+                    Ast::Pending => return Ok(None),
                     Ast::PythonAst(_) => {
-                        if file_info.borrow_mut().file_info_ast.borrow().ast.as_py_ast().indexed_module.is_some() {
-                            return Ok(HoverFeature::hover_python(session, file_symbol, &file_info, params.text_document_position_params.position.line, params.text_document_position_params.position.character));
-                        }
+                        return Ok(HoverFeature::hover_python(session, file_symbol, &file_info, params.text_document_position_params.position.line, params.text_document_position_params.position.character));
                     },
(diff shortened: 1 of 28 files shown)


mmahrouss (approved):
Nice!

server/src/core/odoo.rs
```diff
@@ -1775,22 +1769,22 @@ impl Odoo {
...
                 }
-                if ast_kind == AstKind::PythonAst && file_info.borrow_mut().file_info_ast.borrow().ast.as_py_ast().indexed_module.is_some() {
-                    return Ok(CompletionFeature::autocomplete(session,
-                        file_symbol,
-                        &file_info,
-                        params.context,
-                        params.text_document_position.position.line,
-                        params.text_document_position.position.character
-                    ));
-                }
```
mmahrouss, line 1793: I just saw this `matches!` after the `match` block now, and I was like, we need to merge those, I came here to ask if you want to add it to your PR, 
and you already did! cool:)

PR #727, opened 2026-09-08
[IMP] Add features for compute_sql kwarg for fields

andg-odoo:
Similar to #713 but a little larger as it replicates computes' diagnostics/validation for the parameter needing to exist for OLS03018.


andg-odoo:
Note that since this was added in 19.1+ I'm not sure if all of this should be gated to be checked only in those odoo versions or if it's fine if it's globally accepted.

mmahrouss:
> Note that since this was added in 19.1+ I'm not sure if all of this should be gated to be checked only in those odoo versions or if it's fine if it's globally accepted.

It mainly affects the features, so keeping it won't cause a bug, but I would say it is wrong behavior.

I would perhaps actually add gates on more than 19.1 to be correct, I will specify it in the review

server/src/core/python_arch_eval_hooks.rs
mmahrouss: I would maybe add a version check to >19.1, and only then you can add `compute_sql`. 

it would mean changing the array into a vec, but in this case, it is small enough that it will not matter much

mmahrouss (commented):
Good change, thank you:) I added a comment about the version gating

server/src/core/python_arch_eval_hooks.rs
andg-odoo: Thanks for the review, I've swapped it to a vec and it seems to work well, thank you!

andg-odoo:
@mmahrouss Thanks for the review, updated to vec as well as updated the tests; however, what I've seen in the test cases is that since the context values are determined at startup of odoo-ls(?) it doesn't seem possible to just write to `session.sync_odoo.version = 19.0` in the test case to prove the other side. I could be wrong though.

mmahrouss:
> @mmahrouss Thanks for the review, updated to vec as well as updated the tests; however, what I've seen in the test cases is that since the context values are determined at startup of odoo-ls(?) it doesn't seem possible to just write to `session.sync_odoo.version = 19.0` in the test case to prove the other side. I could be wrong though.

Actually you can, but you are missing to gate also the code `feature_utils.rs` also on version, otherwise, it can never affect `test_get_symbols` because it goes through `find_kwarg_methods_symbols` eventually. 

As for the way you add the test on the version you can check out the test at server/tests/diagnostics/ols03025_to_27.rs

Here is a diff that I used to get it to work: ( it is a quick check, probably badly formatted and has copied code, so you can for sure do better :D )
```diff
diff --git a/server/src/features/features_utils.rs b/server/src/features/features_utils.rs
index 8101fb29..3ed7e95f 100644
--- a/server/src/features/features_utils.rs
+++ b/server/src/features/features_utils.rs
@@ -87,7 +87,11 @@ impl FeaturesUtils {
             if arg_id.as_str() == "inverse_name" {
                 return FeaturesUtils::find_inverse_name_field_symbol(session, from_module, field_value, call_expr);
             }
-            if !["compute", "compute_sql", "inverse", "search"].contains(&arg_id.as_str()){
+
+            if !(
+                ["compute", "inverse", "search"].contains(&arg_id.as_str())
+                || session.sync_odoo.version >= (19, 1) && arg_id.as_str() == "compute_sql"
+            ){
                 return vec![];
             }
         } else {
diff --git a/server/tests/test_get_symbol.rs b/server/tests/test_get_symbol.rs
index a9b30466..b609c200 100644
--- a/server/tests/test_get_symbol.rs
+++ b/server/tests/test_get_symbol.rs
@@ -1,9 +1,13 @@
 ﻿// Test the hover feature by calling get_hover on various symbols in the test addons.
 
+use odoo_ls_server::core::build_scheduler::BuildScheduler;
 use odoo_ls_server::core::odoo::SyncOdoo;
+use odoo_ls_server::core::symbols::SymbolTable;
+use odoo_ls_server::core::symbols::symbol_keys::SourceFileKey;
+use odoo_ls_server::odoo_version::OdooVersion;
 use odoo_ls_server::utils::{PathSanitizer, ToFilePath};
 use odoo_ls_server::Sy;
-use odoo_ls_server::constants::OYarn;
+use odoo_ls_server::constants::{BuildSteps, OYarn};
 use odoo_ls_server::threads::SessionInfo;
 use std::env;
 use std::path::Path;
@@ -11,6 +15,12 @@ use std::path::Path;
 mod setup;
 mod test_utils;
 
+
+fn revalidate(session: &mut SessionInfo, file_sym: SourceFileKey) {
+    SymbolTable::invalidate(session, file_sym, BuildSteps::VALIDATION);
+    BuildScheduler::queue(session, file_sym);
+    BuildScheduler::process_rebuilds(session, false);
+}
 #[test]
 fn test_hover_on_model_field_and_method() {
     // Setup server and session with test addons
@@ -344,15 +354,6 @@ fn test_definition() {
     let range_m2 = file_mgr.borrow().text_range_to_range(&mut session, &module2_test_file, range_m2);
     assert!(compute_kwarg_locs.iter().any(|loc| range_m2 == loc.target_range), "Expected _compute_something to be at the same location as the compute keyword argument in module_2");
 
-    // compute_sql names a method the same way compute does, on the versions that know the kwarg
-    if session.sync_odoo.version >= (19, 1) {
-        let compute_sql_locs = test_utils::get_definition_locs(&mut session, m1_tf_file_symbol, &m1_tf_file_info, 8, 80);
-        assert_eq!(compute_sql_locs.len(), 1, "Expected 1 location for compute_sql method '_compute_something'");
-        let range = session.st().header_range(sym_compute_something[0], Some(m1_tf_file_info.clone())).unwrap();
-        let range = file_mgr.borrow().text_range_to_range(&mut session, &module1_test_file, range);
-        assert_eq!(range, compute_sql_locs[0].target_range, "Expected _compute_something to be at the same location as the compute_sql argument");
-    }
-
     // Now test go to def of `partner_id.country_id.phone_code` on each field.
     let partner_id_locs = test_utils::get_definition_locs(&mut session, m1_tf_file_symbol, &m1_tf_file_info, 33, 25);
     assert_eq!(partner_id_locs.len(), 1, "Expected 1 location for partner_id");
@@ -442,6 +443,19 @@ fn test_definition() {
         "Expected depends' name segment to jump to res.partner's name field, got {:?}",
         depends_name_locs.iter().map(|l| (&l.target_uri, &l.target_range)).collect::<Vec<_>>()
     );
+
+    // compute_sql names a method the same way compute does, on the versions that know the kwarg
+    session.sync_odoo.version = OdooVersion::new(19, 2, 0);
+    revalidate(&mut session, m1_tf_file_symbol);
+    let compute_sql_locs = test_utils::get_definition_locs(&mut session, m1_tf_file_symbol, &m1_tf_file_info, 8, 80);
+    assert_eq!(compute_sql_locs.len(), 1, "Expected 1 location for compute_sql method '_compute_something'");
+    let range = session.st().header_range(sym_compute_something[0], Some(m1_tf_file_info.clone())).unwrap();
+    let range = file_mgr.borrow().text_range_to_range(&mut session, &module1_test_file, range);
+    assert_eq!(range, compute_sql_locs[0].target_range, "Expected _compute_something to be at the same location as the compute_sql argument");
+    session.sync_odoo.version = OdooVersion::new(18, 0, 0);
+    revalidate(&mut session, m1_tf_file_symbol);
+    let compute_sql_locs = test_utils::get_definition_locs(&mut session, m1_tf_file_symbol, &m1_tf_file_info, 8, 80);
+    assert_eq!(compute_sql_locs.len(), 0, "Expected 0 locations for compute_sql method '_compute_something'");
 }
 
 #[test]

```

andg-odoo:
@mmahrouss Thank you, that makes sense. I've gated the parts to 19.1, completion, GoToDe, and diagnostics as well as added tests for each of the cases.

GoToDef and completion appear to respect the version flip at request time not rebuild time so that was simpler. The diagnostic reads the context built at arch eval, so there the test replays a `didChange` after flipping the version, which unloads the symbols and re-runs the field hook, and asserts both sides that way.

I wasn't able to use the nice `# OLS...` comment system for it because that has no concept of a version gate, which is fine, but meant it needed its own fixture and its own file.

mmahrouss:
@andg-odoo Could you please rebase your branch so we can merge it:)

mmahrouss (approved):

andg-odoo:
@mmahrouss Rebased!

PR #728, opened 2026-09-10
[FIX] JS symbol lifecycle

cammarosano:
Before this PR:
JS symbols that at were assets did not support didChange: they were unloaded and never reloaded.
They also did not support creation/rename under module's file tree: they were created as custom entry point.

After: JS symbols under a module's static dir are always assets, and children of the ModuleSymbol.

### DIFF
--- a/server/src/core/odoo.rs
+++ b/server/src/core/odoo.rs
@@ -860,44 +860,70 @@ impl SyncOdoo {
 
     /// Like `unload_path`, but only unloads a symbol when `should_unload` returns true. Returns whether anything was unloaded.
     pub fn unload_path_if(session: &mut SessionInfo, path: &Path, should_unload: impl Fn(&SymbolTable, SymbolKey) -> bool) -> bool {
+        // file/dir might no longer exist in file system, we can't stat it
         let mut unloaded_any = false;
         let ep_mgr = session.sync_odoo.entry_point_mgr.clone();
         for entry in ep_mgr.borrow().iter_all() {
-            let path_str = path.sanitize_cow();
-            let sym_in_data = entry.borrow().data_file_symbols.get(path_str.as_ref()).copied();
-            if let Some(sym) = sym_in_data {
-                if let Some(sym) = sym.upgrade(session.st())
-                    && should_unload(session.st(), sym.into()) {
-                        SymbolTable::unload(session, sym.into());
-                        unloaded_any = true;
-                    }
-                continue;
-            }
-            let sym_in_js = entry.borrow().js_symbols.get(path_str.as_ref()).cloned();
-            if let Some(sym) = sym_in_js {
-                if let Some(sym) = sym.upgrade(session.st())
-                    && should_unload(session.st(), sym.into()) {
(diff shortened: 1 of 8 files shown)


server/src/core/symbols/manifest_arch_eval.rs
```diff
@@ -105,50 +109,54 @@ impl ModuleSymbol {
...
+
+    /// The module owning `path` as an asset, and the files to load for it. `path` may be a
+    /// directory, which a folder rename delivers as a single event.
+    fn asset_owner(session: &SessionInfo, path: &Path) -> Option<(ModuleKey, Vec<PathBuf>)> {
+        if session.sync_odoo.config.is_javascript_disabled() {
+            return None;
+        }
+        let path_str = path.sanitize();
+        let module = js_module_scope::module_of_path(session, &path_str)?;
+        let is_lib = asset_folder_of(&session.st()[module].path, &path_str)? == "lib";
```
mmahrouss:
```suggestion
        let path_str = path.sanitize_cow();
        let module = js_module_scope::module_of_path(session, path_str.as_ref())?;
        let is_lib = asset_folder_of(&session.st()[module].path, path_str.as_ref())? == "lib";
```

server/src/core/pre_parser.rs
```diff
@@ -314,32 +280,20 @@ fn pre_parse_module(ctx: &WorkerCtx, job: Job) {
...
-            }
-        }
+    // Pass 3: walk the module's asset folders. Walking here keeps it off the build thread,
+    // which finds the result memoized when it reaches this module. A module only ever owns
+    // the files under its own `static/`, so no two jobs prepare the same file.
+    if ctx.javascript_disabled {
+        return;
     }
-
+    let assets = ctx.cache.resolve_assets(&module_path.sanitize());
```
mmahrouss:
```suggestion
    let assets = ctx.cache.resolve_assets(&module_path.sanitize_cow());
```

mmahrouss (approved):
looks good:)

server/src/core/symbols/manifest_arch_eval.rs
cammarosano: yeah! thanks!

cammarosano:
Adapted to review comments, rebased.

server/src/core/odoo.rs
```diff
@@ -860,46 +860,75 @@ impl SyncOdoo {
 
     /// Like `unload_path`, but only unloads a symbol when `should_unload` returns true. Returns whether anything was unloaded.
     pub fn unload_path_if(session: &mut SessionInfo, path: &Path, should_unload: impl Fn(&SymbolTable, SymbolKey) -> bool) -> bool {
+        // file/dir might no longer exist in file system, we can't stat it
+        let is_dir = path.extension().is_none();
```
fda-odoo: A dir name could have an extension as a dir name could have dot, so I don't think this test is valid

fda-odoo (changes requested):
One minor comment, but overall looks good to me! well done !

fda-odoo (approved):

PR #730, opened 2026-09-10
[FIX] address clipply warnings in test

cammarosano:


mmahrouss (approved):
Thank you:) it is a non-functional change so I will merge to not block local precommits

PR #733, opened 2026-09-15
Component Manager

cammarosano:
### :anguished:  Big diff? -> mostly tests and adaptations to new API.
----

Replaces `component_descriptors` and `js_component_by_template` maps with a `ComponentMgr`.

The problems:
- ComponentDescriptors were keyed by class name: two classes with the same name (common across modules) overwrote each other.
- Descriptors were never removed: renamed or deleted classes stayed in the index
  - this messes up component resolution for a given template, needed for hover/definition/completion in xml owl templates
- Super classes were identified by name only:
  - wrong on name collisions
  - does not handle import alias
- component descriptors stored in 2 different places (data duplication): `component_descriptors` map in sync_odoo and in each file's `FileInfo::ast`
- components not searchable by file that declares them:  the whole map had to be iterated matching on file path

The solution: `ComponentMgr`
- single storage place for `ComponentDescriptor`s (slotmap)
- proper clear of old components on js file delete/change
- component identity based on slotmap keys (internal only)
- indexed by file and by template name for fast lookups


server/src/core/js_arch_builder.rs
```diff
@@ -136,6 +140,18 @@ impl<'e> JSArchBuilderVisitor<'e> {
...
+
+    fn super_class(&self, class: &Class) -> Option<SuperClassRef> {
+        let Some(Expression::Identifier(id)) = &class.super_class else {
+            return None
+        };
+        let local_name = id.name.to_string();
+        let class_ref = match self.import_bindings.get(&local_name) {
+            None => SuperClassRef::Local(local_name),
+            Some(imported) => SuperClassRef::Imported(imported.clone())
+        };
```
mmahrouss, line 152: What if the super class is imported and then shadowed? Maybe it is okay to ignore that case, I am just curious

mmahrouss (approved):
Nice, cool addition to the JS toolset, nice, contained and clean:)

server/src/core/js_arch_builder.rs
cammarosano: yeah... indeed this heuristic does not handle that.
I don't think it's worth handling that properly... we would probably have to keep track of the declared variables in the file, taking into account their scope...

mmahrouss: Alrighty, seems like an okay trade-off, but maybe we can highlight that in a comment for the next reader:)

fda-odoo (approved):
That's really good and way more clean !
Looks good to me, you just have to clean the commit messages before merging (and squash?)

server/src/core/js_arch_builder.rs
cammarosano: Done!
https://github.com/odoo/odoo-ls/compare/afeef8effc76047b3ec3bfa40e86ac8ce641b997..c652a5378f8add3a393d4da3bea64c0d9ffc305f

cammarosano:
@fda-odoo  commits have been squashed, rebased into alpha

PR #734, opened 2026-09-16
[FIX] Fix cleanup on file close

mmahrouss:
Fix the conditions on closing files, before this commit, builtin files were wrongly dropped.

Now here we only drop custom or untitled files when closed.

For other external files, we just drop file cache, as was traditionally done in the validation base.

This commit also creates modular functions, such that the checks in validation and on closing are well-reused

Fixes a bug where builtins were dropped as a result of #691


cammarosano (approved):
Thank you!

PR #735, opened 2026-09-16
[FIX] <lambda> in completion results

cammarosano:
Exclude lambda function symbols from `all_members`, so that it doesn't show up in completion results. In fact, they are not really members of the class holding them.

Currently no caller of `all_members` relies on it returning lambda symbols.


mmahrouss (approved):
Thank you:)

PR #737, opened 2026-09-17
[FIX] panic on references CSV header with no ":"

cammarosano:


mmahrouss (approved):
Looks good, thanks for the fix;)

PR #738, opened 2026-09-18
[IMP] Add features for group_expand, selection and init_storage kwargs

andg-odoo:
These field arguments name a model method like compute does, so they get the same completion, go-to-definition and missing-method diagnostic. selection is also read from the first positional argument of Selection and Reference fields. 

- `group_expand`: Existed since v10.0
- `selection`: Existed since 8.0
- `init_storage`: Existed since 19.5 / 20.0


andg-odoo:
Obviously I know 1.6.0 is coming up, I just noticed 3 other methods that was missed, it can go in any release no rush, enjoy OXP!

server/src/features/completion.rs
```diff
@@ -602,6 +602,11 @@ fn complete_call(session: &mut SessionInfo, file: SourceFileKey, expr_call: &ruf
             && SymbolTable::is_specific_field_class(session, callable_sym, &["Many2one", "One2many", "Many2many"]) {
                     return complete_expr(arg, session, file, offset, is_param, &[ExpectedType::MODEL_NAME]);
             }
+            if callable_sym.typ() == SymType::CLASS
+            && arg_index == 0
+            && SymbolTable::is_specific_field_class(session, callable_sym, &["Selection", "Reference"]) {
+                    return complete_expr(arg, session, file, offset, is_param, &[ExpectedType::METHOD_NAME]);
+            }
```
mmahrouss, line 610:
```suggestion
            if callable_sym.typ() == SymType::CLASS
                && arg_index == 0
            {
                // Relational fields first argument is a model name.
                if SymbolTable::is_specific_field_class(session, callable_sym, &["Many2one", "One2many", "Many2many"]) {
                    return complete_expr(arg, session, file, offset, is_param, &[ExpectedType::MODEL_NAME]);
                }
                // Selection and Reference fields first argument is a method name.
                if SymbolTable::is_specific_field_class(session, callable_sym, &["Selection", "Reference"]) {
                    return complete_expr(arg, session, file, offset, is_param, &[ExpectedType::METHOD_NAME]);
                }
            }
```

here we are checking if type is class and arg_index is 0 twice, we could maybe reuse the check like so

mmahrouss (approved):

server/src/features/completion.rs
andg-odoo: Done! Thank you!

andg-odoo:
@mmahrouss Fixed the double check use and pushed!

PR #739, opened 2026-09-18
[FIX] filter out virtual doc artifacts in completions

cammarosano:
tsserver offers some completions in templates that come from the virtual doc itself, but are not really available in owl's evaluation context (so, not really an option for the user).

The worst suggestions are the "__ols_m0" functions, that don't exisit anywhere in the "real" code.

image


server/tests/test_js_owl_features.rs
```diff
@@ -215,6 +215,15 @@ fn test_completion_in_template(session: &mut SessionInfo, fixtures: &Fixtures) {
...
         ("shout", CompletionItemKind::METHOD),
     ]);
+    assert_completions(session, xml, "this.shout(|this.props.exclamations)", &[
+        ("Math", CompletionItemKind::VARIABLE)]
+    );
+    assert_no_completions(session, xml, "this.shout(|this.props.exclamations)", &[
+        // virtual doc artifacts
+        "__ols_m0",  // fake function for expression evaluation inside it 
+        "Greeting", // the class imported for typing `this`
+        "arguments" // we see this binding because we evaluation template expression inside fake functions
```
mmahrouss:
```suggestion
        "arguments" // we see this binding because we evaluate template expressions inside fake functions
```

mmahrouss (approved):

fda-odoo (approved):

